PR #8, opened 2019-04-06
Add reference and link to controller availability() in registrar availability()

tayvano:
Helpfully tells people what they should do instead of just telling them they shouldn't do something. 😉😉


Arachnid:
Thank you! Sorry I didn't get to this sooner, my inbox has been a mess since Google Inbox was turned down.

PR #18, opened 2019-05-16
Spelling

briansoule:


Arachnid:
These aren't spelling errors; 'iff' means 'if and only if'.

PR #25, opened 2019-10-25
Add documentation for multicoin support

makoto:


contract-api-reference/publicresolver.md
```diff
@@ -64,6 +65,44 @@ Emits the following event:
...
+
+```text
+function addr(bytes32 node, uint coinType) external view returns(bytes memory);
+```
+
+Returns the Blockchain address associated with the provided `node` and `cointType`, or 0 if none.
+
+This function has interface ID _0xf1cb7e06_.
+
+This function is specified in [EIP 137](https://eips.ethereum.org/EIPS/eip-137).
```
Arachnid: Can you update this to reflect the correct EIP?

contract-api-reference/publicresolver.md
```diff
@@ -64,6 +65,44 @@ Emits the following event:
...
+
+To convert the binary representation into the address, use `formatsByCoinType[0].encoder(binary)` of  [address-encoder](https://github.com/ensdomains/address-encoder)
```
Arachnid: A slightly more verbose description of the format here would be good - you could lift it from the EIP.

makoto: Done

contract-api-reference/publicresolver.md
```diff
@@ -64,6 +65,49 @@ Emits the following event:
...
+
+Returns the Blockchain address associated with the provided `node` and `cointType`, or 0 if none.
+
+This function has interface ID _0xf1cb7e06_.
+
+This function is specified in [EIP 2304](https://eips.ethereum.org/EIPS/eip-2304).
+
+The return value is the cryptocurency address in its native binary format and each blockchain address has a different encoding and decoding method.
+
+For example, the Bitcoin address `1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa` base58check decodes to the 21 bytes `0062e907b15cbf27d5425399ebf6f0fb50ebb88f18` whereas the BNB address `bnb1grpf0955h0ykzq3ar5nmum7y6gdfl6lxfn46h2` Bech32 decodes to the binary representation `40c2979694bbc961023d1d27be6fc4d21a9febe6`.
```
Arachnid: This leaves out the crucial step of encoding as a scriptPubKey, and will make people think that this is the binary representation they should expect.

makoto: Can you put the exact phrase to be placed in this section? I do't think people will understand what base58check or scriptPubKey are unless they read the entire "Address encoding" section of the EIP.

Arachnid: Just replace the example encoded data with the final output from the EIP.

Arachnid: Just replace the example encoded data with the final output from the EIP.

makoto: Done

PR #57, opened 2021-06-02
Replace vnd. with com.

sean-b765:
To reflect changes required for issue https://github.com/ensdomains/ens-app/issues/1140

### DIFF
--- a/contract-api-reference/publicresolver.md
+++ b/contract-api-reference/publicresolver.md
@@ -249,12 +249,12 @@ Standard values for `key` are:
 | notice | A notice regarding this name |
 | keywords | A list of comma-separated keywords, ordered by most significant first; clients that interpresent this field may choose a threshold beyond which to ignore |
 
-In addition, anyone may specify vendor-specific keys, which must be prefixed with `vnd.`.The following vendor-specific keys are currently known:
+In addition, anyone may specify vendor-specific keys, which must be prefixed with `com.`.The following vendor-specific keys are currently known:
 
 | key | Meaning |
 | :--- | :--- |
-| vnd.twitter | Twitter handle |
-| vnd.github | Github username |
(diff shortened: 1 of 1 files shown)


contract-api-reference/publicresolver.md
```diff
@@ -249,12 +249,12 @@ Standard values for `key` are:
 | notice | A notice regarding this name |
 | keywords | A list of comma-separated keywords, ordered by most significant first; clients that interpresent this field may choose a threshold beyond which to ignore |
 
-In addition, anyone may specify vendor-specific keys, which must be prefixed with `vnd.`.The following vendor-specific keys are currently known:
```
Arachnid, line 252: This should remain as-is

Arachnid (changes requested):
Thank you!

contract-api-reference/publicresolver.md
sean-b765: Oh okay, I'm still new to this all, but to me it seems a bit confusing now that one sentence says to use vnd, and then goes on to say 'The following keys are known: com.github, com.twitter...' Plus I thought the keys were being renamed from vnd to com?

Arachnid: You're right; it needs a bigger rewrite.

The long and the short of it is that `vnd.` still means 'vendor specific'. But, we're no longer using vendor-specific prefixes for GitHub and Twitter, we're using their domain names instead. So the stuff about vnd. should remain the same, but GitHub and Twitter should be in a list of known keys that aren't vendor-specific.

sean-b765: OK I get you, if I had more info I'd rewrite this section but I'm not sure of some things still. Thanks for clarifying though!

Arachnid (approved):

PR #73, opened 2021-12-06
normalize in eth-ens-namehash doesn't use Unicode

tlammens:
The forked eth-ens-namehash library at https://www.npmjs.com/package/@ensdomains/eth-ens-namehash does correctly normalize the name using uts46.toUnicode(name, {useStd3ASCII: true, transitional: false}).


Arachnid (approved):
Thanks!

PR #83, opened 2022-01-19
Add Ens l2 offchain section

makoto:


ens-improvement-proposals/ens-l2-offchain.md
```diff
@@ -0,0 +1,117 @@
+# ENS offchain data support (including layer 2)
```
Arachnid, line 1: This file is in the wrong directory.

ens-improvement-proposals/ens-l2-offchain.md
```diff
@@ -0,0 +1,117 @@
...
+
+- [EIP-3668: CCIP Read: Secure offchain data retrieval](https://eips.ethereum.org/EIPS/eip-3668)
+- [ENSIP 10: Wildcard Resolution](ens-improvement-proposals/ensip-10-wildcard-resolution)
+- [ENSIP-11: EVM compatible Chain Address Resolution](ens-improvement-proposals/ensip-11-evmchain-address-resolution.md)
+
+**EIP 3668** allows for offchain (including Layer 2/L2) lookups of data in a way that is transparent to clients and provides contract authors to implement whatever validation is necessary; in many cases, this can be provided without any additional trust assumptions over and above those required if data is stored onchain.
+
+**ENSIP 10** is a general way to resolve wildcard (eg: *.foo.eth) on L1. Issuing subdomains and moving the resolution of the parent name offchain allows dapps to create subdomains offchain yet make it accessible through L1.
+
+**ENSIP 11** allows a single name to hold different addresses across multiple EVM compatible chains. Contract addresses can use these EVM chain addresses so that clients can choose appropriate addresses on different chains.
```
Arachnid, line 19: I think this should probably be described separately, as it's not connected directly to the other two.

ens-improvement-proposals/ens-l2-offchain.md
```diff
@@ -0,0 +1,117 @@
...
+
+```js
+import { CCIPReadProvider } from '@chainlink/ethers-ccip-read-provider';
+import { abi as IExtendedResolver_abi } from '@ensdomains/offchain-resolver-contracts/artifacts/contracts/IExtendedResolver.sol/IExtendedResolver.json';
+const IExtendedResolver = new ethers.utils.Interface(IExtendedResolver_abi);
+const baseProvider = ethers.getDefaultProvider(options.provider);
+const provider = new CCIPReadProvider(baseProvider);
+```
+
+The wildcard is still not supported yet, but one can easily be implemented as follows.
```
Arachnid, line 44: I think this snippet is not very useful on its own - especially with 'Resolver' and 'registry' undefined. I don't think we should be encouraging people to resolve names like this.

ens-improvement-proposals/ens-l2-offchain.md
```diff
@@ -0,0 +1,117 @@
...
+
+Please refer to [offchain resolver client example code](https://github.com/ensdomains/offchain-resolver/blob/main/packages/client/src/index.ts#L46) for more detail.
+
+### web3js
+
+Work in progress
+
+### Any other libraries
+
+If you use other libraries or custom integration, you have to add the support on your own. So please read each specification and support accordingly or contact the ENS team for help.
```
Arachnid, line 72: I would lead by saying they should open an issue for support so we can track efforts to add 3668/ENSIP10 support to other libraries.

Arachnid (changes requested):

ens-improvement-proposals/ens-l2-offchain.md
makoto: Moved to FAQ

ens-improvement-proposals/ens-l2-offchain.md
makoto: what is the recommendation? If it's not ready I can just say the support is work in progress

Arachnid: I think that's the best option right now.

Arachnid (approved):

PR #107, opened 2022-08-03
Add ensip-13: SAFE Authentication For ENS

wwhchung:


wwhchung:
The ENSIP of EIP-5131
Uses ENS for secondary wallet authentication via TEXT records

Arachnid (approved):
Wouldn't a simpler implementation be to just have the `mainAddress` have text records listing the other accounts that can sign for it? Generally when verifying a signature you will start with the target account, so discovery isn't a problem here, and you can get the signer from the signed message and look it up to see if it's in the list on `mainAddress`.

Also, you may want to consider simply putting these text records on the reverse record for the name - no need to have a primary ENS name! Or you could support both, checking the reverse record first, then the primary name.

wwhchung:
The reason for it to be in both places is for automatic reverse lookup discovery from the UX perspective. 

If this were to gain widespread adoption, you would want to be able to easily see who you can sign for on the wallet you’re connected as. 

lmk if further context would help

Arachnid:
That makes sense, but you could achieve this by indexing via the subgraph. Avoiding unnecessary transactions seems like a good thing to me.

wwhchung:
The text spec doesn’t seem to have any indication of events being emitted when text records are added or removed. Are there events and if not how would the indexing work?

I also think this is suboptimal because it means that someone could force their wallet on you.

Arachnid:
> The text spec doesn’t seem to have any indication of events being emitted when text records are added or removed. Are there events and if not how would the indexing work?

Good point - it's not specified in the spec, but is implemented in the public resolver: https://github.com/ensdomains/ens-contracts/blob/master/contracts/resolvers/profiles/ITextResolver.sol

> I also think this is suboptimal because it means that someone could force their wallet on you.

Someone could give you signing authority over their wallet without your permission, yes, but I don't think that's a real problem. You could require the text record contain a signature by the target wallet, but I don't think it's necessary.

wwhchung:
The main reason for this be the graph/subgraph is ultimately on chain reverse lookups as you pointed out. Open debate of whether or not it’s worth it vs indexing. 

Kind of similar to whether or not enumerability is supported by 721’s or we expect indexers to be used. To me it comes down to the frequency such functions are used, and signing delegation lookup might end up being very frequent. 

Lmk if you’d like to join the discord discussion. Would love to get ens input and backing.

PR #116, opened 2022-09-22
ENSIP-14 - CCIP metadata

jefflau:


ens-improvement-proposals/ensip-14-ccip-metadata.md
```diff
@@ -0,0 +1,59 @@
...
+
+With CCIP-read subdomains already starting to get used by larger partners in the ENS ecosystem, it is important that there is a way of frontend interfaces to get important metadata to allow a smooth user interface. For instance the owner of a CCIP enabled name would need to be known, for a UI to show whether or not the currently connected account has the rights to edit the records. If the owner was not known, the only way to know would be to make the edit and wait for failure.
+
+This ENSIP addresses this by adding a way of important metadata to be gathered on the offchain resolver, which would likely revert and be also resolved offchain, however there is an option for it to be also left onchain if there value was static and wouldn't need to be changed often.
+
+### Specification
+
+```solidity
+contract OffChainResolver {
+    owner(bytes32 node){
```
makoto, line 26: You should specify the return data type to be an address

makoto:
This protocol also should support CCIP-read so that `recordKeys` can fetch the data from l2. The callback function should not need to implement any verification logic via signature recover nor storage proof as the actual verification should be conducted to the call to fetch the data(eg: addr())

ens-improvement-proposals/ensip-14-ccip-metadata.md
```diff
@@ -0,0 +1,73 @@
...
+
+Add metadata functions to the resolver. The first argument MUST have the node if it is specific to a node.
+
+```solidity
+
+
+interface OffChainResolver {
+    owner(bytes32 node);
+
+    isApprovedForAll(address account, address operator)
```
makoto, line 32: Assuming you have to get this information from offcahain, I am not sure if this is feasible for L2 resolvers.
L1 resolver can verify the storage location of L2 resolver but cannot verify the function call result of L2.

makoto:
Maybe worth mentioning whether each function should get the data from l1 or l2 as an implementation note.

ens-improvement-proposals/ensip-14-ccip-metadata.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Specification
+
+Add metadata functions to the resolver. The first argument MUST have the node if it is specific to a node.
+
+```solidity
+
+
+interface OffChainResolver {
+    owner(bytes32 node);
```
makoto, line 30: Bear in mind that some network (eg: starknet, zksync) don't use the same address format as Ethereum, so better to specify that it returns bytes32 rather than address.

ens-improvement-proposals/ensip-14-ccip-metadata.md
makoto: Same as `owner` but l2 address may not necessarily EVM compatible `address` so `bytes32` may be preferred as both input data type and output data type

ens-improvement-proposals/ensip-14-ccip-metadata.md
```diff
@@ -0,0 +1,73 @@
...
+
+    isApprovedForAll(address account, address operator)
+
+    dataLocation(bytes32 node) returns (string);
+
+    allRecords(bytes32 node) returns (Records);
+
+    recordKeys(bytes32 node) returns (RecordsKeys)
```
TateB, line 38: do we want to specify what the `Records` and `RecordsKeys` structs should be?

makoto: Good idea

How about keeping it same as the graph schema?

Instead of 

```
// {
//   addr: [60, 1]
//   textRecords: ['url', 'avatar']
//   contentHash: true
// }

```

It should be

```
// {
//   coinTypes: [60, 1]
//   texts: ['url', 'avatar']
// }
```

I don't think we need contentHash as you can simply hit the endpoint to figure it out

I don't know what `allRecords` suppose to return so @jefflau needs a bit more clarification.

ens-improvement-proposals/ensip-14-ccip-metadata.md
```diff
@@ -0,0 +1,73 @@
...
+This ENSIP addresses this by adding a way of important metadata to be gathered on the offchain resolver, which would likely revert and be also resolved offchain, however there is an option for it to be also left onchain if there value was static and wouldn't need to be changed often.
+
+### Specification
+
+Add metadata functions to the resolver. The first argument MUST have the node if it is specific to a node.
+
+```solidity
+
+
+interface OffChainResolver {
```
Arachnid, line 29:
```suggestion
interface OffchainResolver {
```

ens-improvement-proposals/ensip-14-ccip-metadata.md
Arachnid, line 30: What do these first two functions return?

ens-improvement-proposals/ensip-14-ccip-metadata.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Motivation
+
+With CCIP-read subdomains already starting to get used by larger partners in the ENS ecosystem, it is important that there is a way of frontend interfaces to get important metadata to allow a smooth user interface. For instance the owner of a CCIP enabled name would need to be known, for a UI to show whether or not the currently connected account has the rights to edit the records. If the owner was not known, the only way to know would be to make the edit and wait for failure.
+
+This ENSIP addresses this by adding a way of important metadata to be gathered on the offchain resolver, which would likely revert and be also resolved offchain, however there is an option for it to be also left onchain if there value was static and wouldn't need to be changed often.
+
+### Specification
+
+Add metadata functions to the resolver. The first argument MUST have the node if it is specific to a node.
```
Arachnid, line 24: I'm not sure I understand the 'MUST' statement here.

ens-improvement-proposals/ensip-14-ccip-metadata.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+interface OffChainResolver {
+    owner(bytes32 node);
+
+    isApprovedForAll(address account, address operator)
+
+    dataLocation(bytes32 node) returns (string);
+
+    allRecords(bytes32 node) returns (Records);
```
Arachnid, line 36: I'm not sure `allRecords` and `recordKeys` are things that the L1 resolver will have access to.

ens-improvement-proposals/ensip-14-ccip-metadata.md
```diff
@@ -0,0 +1,73 @@
...
+
+```solidity
+
+
+interface OffChainResolver {
+    owner(bytes32 node);
+
+    isApprovedForAll(address account, address operator)
+
+    dataLocation(bytes32 node) returns (string);
```
Arachnid, line 34: The return value here needs specifying. It should probably be some kind of URI? Possibly we should also have a method that returns a GraphQL endpoint with a specified schema for reading 'offchain' data.

Arachnid (changes requested):

ens-improvement-proposals/ensip-14-ccip-metadata.md
jefflau: They would be resolved via CCIP-read

ens-improvement-proposals/ensip-14-ccip-metadata.md
jefflau: I've added return types

ens-improvement-proposals/ensip-14-ccip-metadata.md
jefflau: allRecords would return all the records including keys/values that are currently stored on by the gateway.

ens-improvement-proposals/ensip-14-ccip-metadata.md
Arachnid: Can you reword to clarify that?

ens-improvement-proposals/ensip-14-ccip-metadata.md
jefflau: I believe I added this so it would be possible to know which L2 network the data was on, and where to update it directly (this would obviously require a second standard that pairs with ccip-read to create write functionality). This wasn't really meant to be related to reading the data, since we don't need to know where the data is to read it.

ens-improvement-proposals/ensip-14-ccip-metadata.md
makoto: I think @Arachnid also suggested an option to return graphqlurl endpoint so that they can get the data . The alternative approach would be that any data that are not in l1 just throws CCIP error and let the gateway server fetch data via graph.

ens-improvement-proposals/ensip-14-ccip-metadata.md
jefflau: I reworded already, can u check the rewording?

ens-improvement-proposals/ensip-14-ccip-metadata.md
Arachnid: Right. But part of the purpose of this spec is to make it possible for the manager etc to know how to update the data, right?

jefflau: Are you suggesting we use a graphql endpoint to update data? I thought your first comment was about reading

serenae-fansubs:
Should the CCIP metadata be able to include an expiry?

If an offchain name provider is issuing subnames with expiration dates, it would be nice if that expiry could show up in the ENS Manager App too.

makoto:
@serenae-fansubs. Yeah, that's actually a good idea.

AlexNi245:
The proposal for metadata currently assumes that the metadata is limited to a resolver level. 
That leads to the resolver always returning the same metadata regardless of the domain. I think that might limit the capabilities of resolving metadata for subdomains since the resolver is unaware of the domain it should resolve metadata for.

Hence I propose to add the DNS-encoded domain name as a function argument for the metadata function to allow the resolver to return dynamic metadata based on the given domain name.
An example use case might be a top-level domain like "alice.eth" that wants to mount in namespaces from different chains.

So, for example, the subdomain optimism.alice.eth should be resolved by a contract deployed on optimism, whereas "gnosis.alice.eth" should be resolved by a smart contract on the Gnosis chain. 

To retrieve metadata for both optimism.alice.eth and gnosis.alice.eth the metadata function needs to be aware of the name it's supposed to resolve metadata for.  

For that reason, I think it makes sense to pass the dnsEncoded name through the metadata function. 

``` solidity
function metadata(bytes calldata name) external view returns (string memory, uint256, string memory, uint8, bytes memory) {
        (CcipVerifier memory _ccipVerifier, ) = getVerifierOfDomain(name);
        return ICcipResponseVerifier(_ccipVerifier.verifierAddress).metadata(name);
    }

File: ens-improvement-proposals/ensip-16-offchain-metadata.md [hunk 9]
```diff
@@ -0,0 +1,167 @@
...
+const resolver = await ens.resolver(node)
+const owner = await resolver.owner(node)
+// 0x123...
+const dataLocation = await.resolver.graphqlUrl()
+// {
+//   url: 'http://example.com/ens/graphql',
+// }
+```
+
+##### GgraphQL schema
```
jefflau, line 92:
```suggestion
##### GraphQL schema
```

ens-improvement-proposals/ensip-16-offchain-metadata.md
```diff
@@ -0,0 +1,167 @@
...
+type Domain {
+  id: ID! # concatenation of context and namehash delimited by `-`
+  context: Bytes
+  name: String
+  namehash: Bytes
+  labelName: String
+  labelhash: Bytes
+  resolvedAddress: Bytes
+  parent: Domain
+  subdomains: [Domain]
```
serenae-fansubs, line 139: Can we also add a `subdomainCount` here too?

Also, is pagination (or lack thereof) left up to the implementer to support or not?

ens-improvement-proposals/ensip-16-offchain-metadata.md
```diff
@@ -0,0 +1,168 @@
+---
+description: Allows metadata to be queried on ERC-3668 enabled names
+---
+
+# ENSIP-16: Offchain metadata
+
+| **Author**    | Jeff Lau \<jeff@ens.domains> |
```
Arachnid, line 7: @makoto You should add yourself here too

ens-improvement-proposals/ensip-16-offchain-metadata.md
```diff
@@ -0,0 +1,168 @@
...
+| **Status**    | Draft                        |
+| **Submitted** | 2022-09-22                   |
+
+### Abstract
+
+This ENSIP allows metadata to be queried directly on the resolver for ERC-3668(CCIP Read: Secure offchain data retrieval) enabled names. ERC-3668 will power many of the domains in the future, however since the retrieval mechanism uses wildcard + offchain resolver, there is no standardised way to retrieve important metadata information such as the owner (who can change the records), or which L2/offchain database the records are stored on.
+
+### Motivation
+
+With ERC-3668 subdomains already starting to get used by larger partners in the ENS ecosystem, it is important that there is a way of frontend interfaces to get important metadata to allow a smooth user interface. For instance the owner of a ERC-3668 enabled name would need to be known, for a UI to show whether or not the currently connected account has the rights to edit the records. If the owner was not known, the only way to know would be to make the edit and wait for failure.
```
Arachnid, line 18:
```suggestion
With ERC-3668 subdomains already starting to see wide adoption, it is important that there is a way for frontend interfaces to get important metadata to allow a smooth user experience. For instance a UI needs to be able to check if the currently connected user has the right to update an EIP-3668 name.
```

ens-improvement-proposals/ensip-16-offchain-metadata.md
```diff
@@ -0,0 +1,168 @@
...
+```solidity
+
+interface OffChainResolver {
+    /** @dev Returns the owner of the resolver on L2
+     * @param node
+     * @return owner in bytes32 instead of address to cater for non EVM based owner information
+     */
+    owner(bytes32 node) returns (bytes32 owner);
+
+    isApprovedForAll(address account, address operator) returns (boolean);
```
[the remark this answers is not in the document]
Arachnid: We should specify two separate interfaces - one basic one that all L1 CCIP-Read enabled resolvers should implement, and one L2 one for ownership data etc that can be implemented on EVM-compatible L2s.

ens-improvement-proposals/ensip-16-offchain-metadata.md
```diff
@@ -0,0 +1,168 @@
...
+const resolver = await ens.resolver(node)
+const owner = await resolver.owner(node)
+// 0x123...
+const dataLocation = await.resolver.graphqlUrl()
+// {
+//   url: 'http://example.com/ens/graphql',
+// }
+```
+
+##### GraphQL schema
```
Arachnid, line 92: This needs a short description as to why it's being specified and how it will operate.

Arachnid (changes requested):

ens-improvement-proposals/ensip-16-offchain-metadata.md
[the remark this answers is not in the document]
makoto: `one basic one that all L1 CCIP-Read enabled resolvers should implement,` this is basically what `owner` and `isApprovedForAll` on L1 resolver can implement as long as it also takes `context` as an argument

ens-improvement-proposals/ensip-16-offchain-metadata.md
Arachnid: I had in mind something more specific to this EIP, like "this specifies the schema for a standardized metadata subgraph, which collects information emitted by L1 and L2 resolvers".

Arachnid (approved):

makoto (approved):

PR #120, opened 2022-10-24
Adding a quick testnet migration doc by Alchemy

johnfawole:
Hi maintainers,

Thanks for accepting the earlier commit on Goerli. We felt it would be great if the devs reading the docs have handy content to switch to Goerli.

Therefore, we added one link to migrate from Ropsten to Goerli. We believe this would not only be helpful but would also enhance developers' experience.

Regards,
John,
An Ambassador at Alchemy.


Arachnid (approved):

johnfawole:
ENS has merged the PR with the Alchemy link.

PR #123, opened 2023-01-06
Typo fix: changed "nonfungable" to "non-fungible"

archie-mckenzie:
In eth-permanent-registrar, see: https://docs.ens.domains/contract-api-reference/.

Changed the typo "nonfungable" to correct form "non-fungible", as found elsewhere in the docs.


lucemans (approved):
LGTM

PR #132, opened 2023-03-14
Evm-chain reverse resolution

jefflau:


ens-improvement-proposals/ensip-16-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,35 @@
...
+
+This ENSIP specifies a way for reverse resolution to be used on other EVM chains.
+
+## Motivation
+
+Reverse resolution has been in use since ENS's inception, however at the time Ethereum had no concrete scaling plans. In the past 5 years, we've seen layer 2s and sidechains become more prevalent and we first allowed support for these with ENSIP-9 (formerly EIP-2304) to allow addresses from other chains to be stored on ENS. Reverse resolution can be expanded to allow the forward resolution to first check the records for the chain that the user is on, before falling back to using the address record on mainnet.
+
+## Specification
+
+* Use ENSIP-3 to reverse resolve an address to a name.
```
makoto, line 23: ENSIP-1 doesn't specify how it does reverse lookup  for multi chain and storing all the data on L1 is not feasible if we want to support contract base account (eg: multisig wallet), because L1 has no knowledge of who the signers of the multisig on other chains nor be able to call `isValidSignature` specified by EIP-1271 on chain.

The alternative approach is store the reverse record in each l2 chain and have l1 reverse record `.reverse` to have chain specific namespace (eg: `.[[coinType](url)].reverse`) that fetches the data via L1 using CCIP-read.

NOTE: coinType follows [ENSIP-11](https://docs.ens.domains/ens-improvement-proposals/ensip-11-evmchain-address-resolution)

### example use case

1. ENS lab creates a multisg contract 0x01 on Arbitrum
2. ENS lab sets L2(Arbitrum) reverse registrar as `0x01.arbitrumcointype.reverse=arbitrum.ens.eth` on Arbitrum
3. When a dapp wants to display the primary name of 0x01, the site lookup 0x01.aribtrumcointype.reverse on Mainnet but fetches the data via Arbitrum through CCIP-read

 The only modification to client code is that the dapp has to passes the chain info (aribtrumcointype) in the lookup at #3, but at least users don’t have to switch networks for the read operation.

jefflau: I have written a little bit about this in this gist: https://gist.github.com/jefflau/7e9d303078891a969bfe70011e347bc0#canonical-registry-on-each-supporting-chain-with-discovery-mechanism

I did not specify the discovery mechanism, but I think having it using ccip-read is a reasonable option. The other alternative I could think of is using a resolver and using the addr record on the resolver. But I think using ccip-read makes a lot of sense here is probably a superior option

PR #133, opened 2023-03-20
ENSIP-6 Improvements

buffrr:
This PR updates the status of ENSIP-6 from Stagnant to Draft and enhances compatibility with DNS without requiring changes to the DNS resolver profile. Instead, it establishes a better convention for translating ENS nodes and record sets to DNS.

The main improvement involves treating record sets as part of the top-level domain zone which may work better for .eth and enables several useful features:

- Nodes can delegate to a standard nameserver without coordinating with parent nodes by simply adding an NS record set.
- DNSSEC support is enabled through the use of DS record sets, effectively turning DNS into a secure layer 2 solution.
- CNAMEs can be added at the domain level, since domains are not at the zone apex (e.g., `(nick.eth, nick.eth, CNAME)`).
- Establishes a clearer relationship between record sets and ENS nodes.

This approach makes it theoretically possible to export all ENS nodes implementing this ENSIP into a DNS-compliant zone that can be consumed by standard DNS software with support for zone transfers ... etc.

There are other ways to do this, most come with their own trade-offs. This proposal does not require changing any contracts, but instead provides a clearer convention for those wanting to build DNS integrations for ENS but of course I'm open to other ideas. Thanks!

### DIFF
--- a/ens-improvement-proposals/ensip-6-dns-in-ens.md
+++ b/ens-improvement-proposals/ensip-6-dns-in-ens.md
@@ -6,13 +6,10 @@ description: >-
 
 # ENSIP-6: DNS-in-ENS
 
-{% hint style="danger" %}
-This standard is obsolete, and does not reflect the current implementation.
-{% endhint %}
 
-| **Author**  | Jim McDonald (@mcdee) |
+| **Author**  | Jim McDonald (@mcdee), Hasan Adams (@buffrr) |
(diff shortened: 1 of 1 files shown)


gskril:
Sorry this never got a solid review. Closing it for now becuase ENSIP's have since moved to https://github.com/ensdomains/ensips. If you'd like to revive the proposal, please re-open it there!

PR #141, opened 2023-05-02
ENSIP-16: Proposal for "meta-resolution" for off-chain names.

serenae-fansubs:


Arachnid:
Should this be combined with https://github.com/ensdomains/docs/pull/116 ?

PR #144, opened 2023-06-09
Fix scroll location tag for spec.json description section

fselmo:
Pretty minor fix but something I noticed as I was scrolling through the info for ENSIP-15. 

The section heading actually uses a "." for `#description-of-spec.json` so this link was broken for me on Firefox.


fselmo:
Stale

PR #148, opened 2023-07-05
Update writing-a-registrar.md

naftalimurgor:
## Summary

- minor improvments eg: `address(0x0)` comparison for address `0x0` check
- switch to type `address` since `var` is long deprecated


naftalimurgor:
Hey @makoto 
What's your take on this, I saw it would be great to do this small change(and other future changes) to keep the docs updated
Thanks

gskril:
Sorry we didn't get around to addressing this PR sooner, but our subname docs should be much better these days ([this page](https://docs.ens.domains/web/subdomains) is an entry point). Always open to contributions, I'm actively watching this repo :)

PR #151, opened 2023-07-20
Dynamic Metadata

AlexNi245:
At the moment metadata is not aware of the domain it should return metadata for. 
This PR proposes a solution to pass the domain name through the metadata function enabling a more flexible use of metadata.

### DIFF
--- a/ens-improvement-proposals/ensip-15-ccip-read-metadata.md
+++ b/ens-improvement-proposals/ensip-15-ccip-read-metadata.md
@@ -1,0 +1,137 @@
+---
+description: Allows metadata to be queried on CCIP enabled names
+---
+
+# ENSIP-XX: CCIP-read metadata
+
+| **Author**    | Jeff Lau \<jeff@ens.domains> |
+| ------------- | ---------------------------- |
+| **Status**    | Draft                        |
+| **Submitted** | 2022-09-22                   |
+
+### Abstract
+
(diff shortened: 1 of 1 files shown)


ens-improvement-proposals/ensip-15-ccip-read-metadata.md
```diff
@@ -0,0 +1,137 @@
...
+     */
+    owner(bytes32 node) returns (bytes32 owner);
+
+    isApprovedForAll(address account, address operator) returns (boolean);
+    /** @dev Returns the owner of the resolver on L2
+     * @return name can be l2 chain name or url if offchain
+     * @return coinType according to https://github.com/ensdomains/address-encoder
+     * @return graphqlUrl url of graphql endpoint that provides additional information about the offchain name and its subdomains
+     * @return storageType 0 = EVM, 1 = Non blockchain, 2 = Starknet
+     * @return context can be l2 resolver contract address for evm chain but can be any l2 storage identifier for non evm chain
```
makoto, line 45: I had a discussion with @Arachnid and would like to change the following

- @return context = This returns the same data used as `context` when calling `addr(bytes calldata context, bytes32 node, uint256 coinType)` at https://github.com/corpus-io/ENS-Bedrock-Resolver/blob/main/contracts/l2/profiles/addr/AddrResolver.sol#L98 (in your specific implementation, it needs to return the owner address of the name).
- @return storageLocation = This is where it returns the l2 resolver contract address for evm chain but can be any l2 storage identifier for non evm chain

makoto: @AlexNi245 can you add storage location? Also can you recreate this PR against `ensip-14` branch rather than against `master`?

PR #155, opened 2023-08-07
Update links and contract deployments

gskril:
Different types of contracts used to live in their own repos. At some point they were consolidated into [ens-contracts](https://github.com/ensdomains/ens-contracts) but the docs weren't updated. This should bring most pages up to date.

Remaining questions:
- Is there a better example for [Writing a Registrar](https://docs.ens.domains/contract-developer-guide/writing-a-registrar) besides [this repo](https://github.com/ensdomains/subdomain-registrar)?
- Some ENSIP's [like this](https://docs.ens.domains/ens-improvement-proposals/ensip-7-contenthash-field#implementation) link to old repos - should we also update those links, or leave them for the sake of having standards use original content?


contract-api-reference/.eth-permanent-registrar/README.md
```diff
@@ -14,7 +14,7 @@ The registrar itself is called [BaseRegistrar](registrar.md). This contract impl
 * Name owners may reclaim ownership in the ENS registry if they have lost it.
 * Owners of names in the legacy registrar may transfer them to the new registrar, during the 1 year transition period. When they do so, their deposit is returned to them in its entirety.
 
-In addition, the registrar is an [ERC721](https://github.com/ensdomains/ens/blob/master/docs/ethregistrar.rst#id3) compliant nonfungable token contract, meaning that .eth registrations can be transferred in the same fashion as other NFTs.
+In addition, the registrar is an ERC721 compliant nonfungable token contract, meaning that .eth registrations can be transferred in the same fashion as other NFTs.
```
makoto, line 17: You may also want to mention about NameWrapper which is ERC1155

dapp-developer-guide/ens-as-nft.md
```diff
@@ -1,6 +1,6 @@
 # ENS as NFT
 
-When ENS .eth registrar migrated in May 2019, the .eth registrar became an [ERC721](https://github.com/ensdomains/ens/blob/master/docs/ethregistrar.rst#id3) compliant non-fungible token contract, meaning that .eth registrations can be transferred in the same fashion as other NFTs.
+When ENS .eth registrar migrated in May 2019, the .eth registrar became an ERC721 compliant non-fungible token contract, meaning that .eth registrations can be transferred in the same fashion as other NFTs.
```
makoto, line 3: Same here. Please also add a sentence about NameWrapper which is ERC721

deploying-ens-on-a-private-chain.md
```diff
@@ -75,7 +75,7 @@ Above, we first create a new top-level domain, "resolver", then set its resolver
...
+So far, domains can only be registered manually by the owner of the registry's root node. Fortunately, contracts can also own nodes. This means we can set up a registrar contract as the owner of a node, e.g. "test", in the registry which enables it to distribute subdomains such as "mycontract.test". It allows us to have custom, on-chain logic which governs domain allocation. Once we own a \(sub-\)node we are free to repeat this process and set up another registrar. If you are part of the "myorg" organisation you could register "myorg.test" and let it point to your custom registrar which only allows certified members of your organisation to claim subdomains such as "bob.myorg.test". For our private network, we'll use the simple 'first come, first served' [FIFSRegistrar](https://github.com/ensdomains/ens-contracts/blob/master/contracts/registry/FIFSRegistrar.sol), and set it as the owner of the top-level domain "test" in our migration script:
```
makoto, line 79: This whole section may be out of date. You may simply point to https://github.com/ensdomains/ensjs-v3/tree/main/packages/ens-test-env that comes with [deployment scripts](https://github.com/ensdomains/ensjs-v3/tree/main/packages/ensjs/deploy) or @jefflau should review the code given he recently did testnet deployment

contract-api-reference/.eth-permanent-registrar/README.md
gskril: But technically the ERC1155 isn't in the .eth permanent registrar, right?

My understanding is that a name's ERC721 stays in the .eth registrar owned by the NameWrapper, then the NameWrapper mints a new ERC1155 to represent ownership.

deploying-ens-on-a-private-chain.md
jefflau: I have added an internal doc to Leon's new doc. We could add it here, but it would likely need some explanation and expanding, but works as an internal doc for now

jefflau:
I have a subdomain registrar PR on ens-contracts, but it is still incomplete. It's definitely a more complicated registrar, but it depends what we mean by 'better' example.

deploying-ens-on-a-private-chain.md
gskril: I'm just going to keep it as-is for now - it's not any worse than it was before and it sounds like we'll have a major upgrade soon, so doesn't seem like a blocker to the other changes in this PR.

PR #157, opened 2023-08-24
Evm reverse resolution

jefflau:
Renamed the branch so it isn't ensip-16. Old comments/PR here: https://github.com/ensdomains/docs/pull/132


ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,75 @@
...
+* User can now claim their reverse on this L2 and set it to their ENS name
+* Their ENS name will need to setup their record for the same EVM-cointype as the network, which is specified in [ENSIP-11](https://docs.ens.domains/ens-improvement-proposals/ensip-9-multichain-address-resolution)
+* A dapp can then detect the chainID that a user is on, find the corresponding cointype and resolve their primary ens name by resolving the name record at [userAddress].[evmChainCointype].reverse. This will be resolved via CCIP-read and look up the reverse record on the corresponding EVM-chain.
+* Dapp will then resolve this name via ENS on L1 to check if the forward resolution matches. This forward resolution can be on L1, or the user can set up CCIP-read records for each network and put those addresses wherever they want.
+
+### Resolving Primary ENS Names by a dapp
+
+* If a dapp has a connected user it SHOULD first check the chainId of the user
+* If the chainId is not 1, it SHOULD then construct the ENS name [address].[evmChainCoinType].reverse to obtain the primary ENS name of the user. If none is found, it SHOULD assume that the user has no primary ENS name set
```
moldy530, line 37: this is cool and kinda what I had in mind but I like the proposal of having L2 registrars in the mix, still grokking but 👍

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,75 @@
...
+
+With account abstraction becoming more popular, it is becoming increasingly important to be able to setup different addresses for each L2, but using the same ENS name. It is no longer good practise to assume that the address of a user in mainnet, is also controlled by the same user on an EVM-compatible L2. This can be solved with ENS names resolving to different addresses on each chain, but also additionally the primary ENS name being set to the same ENS name on each chain preserving the identity of the user across chains.
+
+## Specification
+
+### Overview
+
+* Reverse registrars will be setup on each L2, with a corresponding registry
+* Reverse registrar will only allow setting the name, without resolver customisability. This is to allow a consistent storage slot that can be checked on L1.
```
moldy530, line 28: does this mean that we can't resolve via CCIP-read on L2 to defer storage to an off-chain DB?

jefflau: That's correct. Reverse names are already using CCIP-read to resolve from L1 for a specific L2. But the reverse record itself is required to be on-chain (on the L2) so it can be trustless proven on L1

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
jefflau: Appreciate any and all feedback!

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
moldy530: How would this work for subdomains registered via CCIP read on L1?

The example that comes to mind is `cb.id`. If I have `moldy.cb.id`, then my addresses are mapped on CB servers. But if CB wants to also expose the per chain reverse mapping, they'd have to do the registration on the L2 and pay that gas?

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Being explicit about default Primary ENS Name
+
+To make things explicit we will allow the signing of a message to confirm that the address in question would like to use mainnet or another network as fallback. This would either resolve directly on mainnet. Defaults would only be applicable to EoAs that can sign a message. This is because smart contract accounts would not be able to reliably set a default on all chains.
+### Setting default
+
+1) Sign a message to set a default record
+2) call `setName()` on the default registrar on L1
+
+Possibility: Allowing L1 Primary ENS names to also use `default.addr`. This could be incorporated into the public resolver's `name()` function.
```
makoto, line 78: Should this be `default.reverse` ?

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
makoto: That's correct for the time being. In future we should consider moving the default namespace to more gas cost efficient storage option.

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,87 @@
...
+## Specification
+
+### Overview
+
+* Reverse registrars will be setup on each L2, with a corresponding registry
+* Reverse registrar will only allow setting the name, without resolver customisability. This is to allow a consistent storage slot that can be checked on L1.
+* User can now claim their reverse on this L2 and set it to their ENS name
+* Their ENS name will need to setup their record for the same EVM-cointype as the network, which is specified in [ENSIP-11](https://docs.ens.domains/ens-improvement-proposals/ensip-9-multichain-address-resolution)
+* A dapp can then detect the chainID that a user is on, find the corresponding cointype and resolve their primary ens name by resolving the name record at [userAddress].[evmChainCointype].reverse. This will be resolved via CCIP-read and look up the reverse record on the corresponding EVM-chain.
+* Dapp will then resolve this name via ENS on L1 to check if the forward resolution matches. This forward resolution can be on L1, or the user can set up CCIP-read records for each network and put those addresses wherever they want.
```
Arachnid, line 32:
```suggestion
* Reverse registrars will be setup on each L2, with a corresponding registry.
* Reverse registrar will only allow setting the name, without resolver customisability. This is to allow a consistent storage slot that can be checked on L1.
* User can now claim their reverse on this L2 and set it to their ENS name.
* Their ENS name will need to set their record for the same EVM-cointype as the network, which is specified in [ENSIP-11](https://docs.ens.domains/ens-improvement-proposals/ensip-9-multichain-address-resolution).
* A dapp can then detect the chainID that a user is on, find the corresponding cointype and resolve their primary ens name by resolving the name record at [userAddress].[evmChainCointype].reverse. This will be resolved via CCIP-read and look up the reverse record on the corresponding EVM-chain.
* Dapp will then resolve this name via ENS on L1 to check if the forward resolution matches. This forward resolution can be on L1, or the user can set up CCIP-read records for each network and put those addresses wherever they want.
```

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Resolving Primary ENS Names by a dapp
+
+1) If a dapp has a connected user it SHOULD first check the chainId of the user
+2) If the chainId is not 1, it SHOULD then construct the ENS name [address].[evmChainCoinType].reverse to obtain the primary ENS name of the user. 
+3) If none is found, it SHOULD check check mainnet [address].default.reverse
+4) If the dapp finds an ENS name, it MUST first check the forward resolution. The forward resolution MUST match the same coin type as the chain id that the user.
```
Arachnid, line 39:
```suggestion
1) If a dapp has a connected user it SHOULD first check the chainId of the user.
2) If the chainId is not 1, it SHOULD then construct the ENS name [address].[evmChainCoinType].reverse to obtain the primary ENS name of the user. 
3) If none is found, it SHOULD check check mainnet [address].default.reverse.
4) If the dapp finds an ENS name, it MUST first check the forward resolution. The forward resolution MUST match the same coin type as the chain id that the user.
```

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,87 @@
...
+
+Note: The dapp MUST NOT use mainnet even if the Primary ENS name has not been set on that chain or as a default, and must instead show the address.
```
Arachnid, line 41:
```suggestion
Note: The dapp MUST NOT use the reverse record set for mainnet even if the Primary ENS name has not been set on the target chain, and must instead show the address.
```

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Resolving an avatar by a dapp on L2
+
+ENSIP-12 was concieved before the ENS L2 reverse resolution specification and therefore should be updated to reflect the current state of ENS primary name resolution. This means that all avatar records are able to be updated on a per-chain basis by updating the avatar record on their reverse node.
```
Arachnid, line 45: This seems to conflict with the summary saying users can only set the name.

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,87 @@
...
+#### Example for an EVM Address
+
+To determine the avatar URI for a specific EVM chain address, the client MUST reverse-resolve the address by querying the ENS registry on Ethereum for the resolver of `<address>.<evmChainId>.reverse`, where `<address>` is the lowercase hex-encoded Ethereum address, without leading '0x'. Then, the client calls `.text(namehash('<address>.<evmChainId>.reverse'), 'avatar')` to retrieve the avatar URI for the address.
+
+If a resolver is returned for the reverse record, but calling `text` causes a revert or returns an empty string, the client MUST call `.name(namehash('<address>.<evmChainId>.reverse'))`. If this method returns a valid ENS name, the client MUST:
+
+1. Validate that the reverse record is valid, by resolving the returned name and calling `addr` on the resolver, checking it matches the original <chaiId> address.
+2. Perform the process described under 'ENS Name' to look for a valid avatar URI on the name.
+
+A failure at any step of this process MUST be treated by the client identically as a failure to find a valid avatar URI.
```
Arachnid, line 56: Probably simpler to say that it should perform a reverse resolution and then look up the avatar on the primary name. You can copy the text from the ENSIP that describes avatar resolution for this.

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,87 @@
...
+
+1) Sign a message to set a default record
+2) call `setName()` on the default registrar on L1
+
+Possibility: Allowing L1 Primary ENS names to also use `default.reverse`. This could be incorporated into the public resolver's `name()` function.
+
+## Further considerations
+
+* Possibility for a resolver that allows gateway choices for the record level to allow a user to split their records between different chains. This would allow them to have arbitrum address records on arbitrum, optimism on optimism and so on.
+* Would require subnames and resolvers setup under [evmChainCoinType].reverse that would need to be governed by the DAO. Each new chain would likely need to be approved by the DAO or could be delegated to a multi-sig at first as things are rolled out and then moved to DAO ownership
```
Arachnid, line 83: This probably doesn't belong here, since it's not relevant to the standard at hand.

Arachnid (commented):
You should probably explicitly include the directions from ENSIP 11 here on how to create a coin type from a chain ID.

Perhaps we should have some syntax for this? Most primary names are going to be for chains that have chain IDs as coin types, and it seems weird that they would all have hugely long decimal identifiers instead of the shorter ones a chain ID would offer. The extended DNS resolver uses the syntax `e<chain id>` as a shorthand; perhaps we can do something like that here too?

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,92 @@
...
+* Their ENS name will need to set their record for the same EVM-cointype as the network, which is specified in [ENSIP-11](https://docs.ens.domains/ensip/11).
+* A dapp can then detect the chainID that a user is on, find the corresponding cointype and resolve their primary ens name by resolving the name record at [userAddress].[coinType].reverse. This will be resolved via CCIP-read and look up the reverse record on the corresponding EVM-chain.
+* Dapp will then resolve this name via ENS on L1 to check if the forward resolution matches. This forward resolution can be on L1, or the user can set up CCIP-read records for each network and put those addresses wherever they want.
+* Once matched, the dapp can then also resolve any text records on the primary ENS name, such as avatar.
+
+### Resolving Primary ENS Names by a dapp
+
+1) If a dapp has a connected user it SHOULD first check the chainId of the user.
+2) If the chainId is not 1, it SHOULD then construct the ENS name [address].[coinType].reverse to obtain the primary ENS name of the user. 
+3) If none is found, it SHOULD check check mainnet [address].default.reverse.
```
Arachnid, line 38: We should clarify behaviour here if `[address].[cointype].reverse` has no resolver vs if it does but does not implement `name`, or returns the empty string.

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,92 @@
...
+4) If the dapp finds an ENS name, it MUST first check the forward resolution. The forward resolution MUST match the same coin type as the chain id of the user.
```
Arachnid, line 39: How does this work for `default.addr`? Will the forward resolver have to have an address set for each chain ID, or do we want a way to set a default address for EVM-compatible chains?

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,92 @@
...
+
+Note: The dapp MUST NOT use the reverse record set for mainnet even if the Primary ENS name has not been set on the target chain, and must instead show the address.
```
Arachnid, line 41:
```suggestion
Note: The dapp MUST NOT use the reverse record set for mainnet even if the Primary ENS name has not been set on the target chain, and must treat this identically to an address with no primary name set.
```

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,92 @@
...
+## Motivation
+
+Reverse resolution has been in use since ENS's inception, however at the time Ethereum had no concrete scaling plans. In the past 5 years, we've seen layer 2s and sidechains become more prevalent and we first allowed support for these with ENSIP-9 (formerly EIP-2304) to allow addresses from other chains to be stored on ENS. Reverse resolution can be expanded to allow the forward resolution to first check the records for the chain that the user is on.
+
+
+## Specification
+
+### Overview
+
+* Reverse registrars will be setup on each L2, with a corresponding registry
```
Arachnid, line 26: We should clarify the deployment and discovery mechanism here. How does the client discover the L2 address of the reverse registrar?

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Resolving an avatar by a dapp on L2
+
+ENSIP-12 was concieved before the ENS L2 reverse resolution specification and therefore should be updated to reflect the current state of ENS primary name resolution. This means that all avatar records are able to be updated on a per-chain basis by updating the avatar record on their reverse node.
+
+#### Example for an EVM Address
+
+To determine the avatar URI for a specific EVM chain address, the client MUST reverse-resolve the address by querying the ENS registry on Ethereum for the resolver of `<address>.<coinType>.reverse`, where `<address>` is the lowercase hex-encoded Ethereum address, without leading '0x'. Then, the client calls `.text(namehash('<address>.<coinType>.reverse'), 'avatar')` to retrieve the avatar URI for the address.
+
+If a resolver is returned for the reverse record, but calling `text` causes a revert or returns an empty string, the client MUST call `.name(namehash('<address>.<coinType>.reverse'))`. If this method returns a valid ENS name, the client MUST:
```
Arachnid, line 51: The way this is worded is overly specific, and implies a method that won't work for a default reverse resolver.

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,92 @@
...
+
+ENS has not been explicit about how to use the mainnet `addr()` record and it is often used as a backup to a user not having an address record set. The mainnet reverse record has also historically been used on other EVM chains due to no alternative on that specific chain. There are a few reasons why it would undesirable to encourage use of mainnet primary name and/or `addr(node, 60)` record as a backup for it not being set on another EVM chain.
+
+An example of why this could be confusing:
+
+Dapp is on Arbitrum and uses mainnet primary ENS name. It resolves the ENS name's mainnet address and uses that to verify the reverse record is correct. It also uses the mainnet address to allow in-app transfers.
+
+Mainnet primary ENS name that has an `addr(node, 60)` that is a smart contract wallet. The smart contract wallet is only on Ethereum and the user in unable to use `CREATE2` to deploy the same smart contract wallet on arbitrum. User 2 sees this Primary ENS name and wants to send funds to User 1. User 2 resolves the `addr()` of the ENS name and sends the funds to an address that doesn't exist on arbitrum and User 1 has no way to access the counterfactual address on that chain.
```
Arachnid, line 75:
```suggestion
Mainnet primary ENS name that has an `addr(node, 60)` that is a smart contract wallet. The smart contract wallet is only on Ethereum and the user is unable to use `CREATE2` to deploy the same smart contract wallet on arbitrum. User 2 sees this Primary ENS name and wants to send funds to User 1. User 2 resolves the `addr()` of the ENS name and sends the funds to an address that doesn't exist on arbitrum and User 1 has no way to access the counterfactual address on that chain.
```

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,92 @@
...
+
+If we mandated that the address cannot use `addr(node, 60)`, but only the address of the chain in question, it would be possible to use mainnet as a backup. However the fact remains that you would still need to claim and set your Primary ENS name on mainnet, and the possibility for confusion seem to outweigh the benefits of using mainnet (high gas) as a catch-all back up for other L2 EVM chains (low gas). Additionally this would only be useful for EVM-compatible chains and would not benefit non-EVM L2s that have a different address format. 
+
+## Being explicit about default Primary ENS Name
+
+To make things explicit we will allow the signing of a message to confirm that the address in question would like to use mainnet or another network as fallback. This would either resolve directly on mainnet. Defaults would only be applicable to EoAs that can sign a message. This is because smart contract accounts would not be able to reliably set a default on all chains.
```
Arachnid, line 81:
```suggestion
To make things explicit we will require the signing of a message to confirm that the address in question would like to use mainnet or another network as fallback. This would either resolve directly on mainnet. Defaults would only be applicable to EoAs that can sign a message. This is because smart contract accounts would not be able to reliably set a default on all chains.
```

Arachnid (changes requested):

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
jefflau: I believe we can't have no resolver in our current architecture because we're putting the L2 reverse registrar together with the name/text records

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
jefflau: Hmm interesting, we never talked about having a default address for EVM-compatible chains, but maybe that is a good idea

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
jefflau: Yeah actually this doesn't really make sense, because there are no custom resolvers on L2 reverse registrars anyway

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
jefflau: I've added a paragraph on this

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,92 @@
...
+
+Reverse resolution has been in use since ENS's inception, however at the time Ethereum had no concrete scaling plans. In the past 5 years, we've seen layer 2s and sidechains become more prevalent and we first allowed support for these with ENSIP-9 (formerly EIP-2304) to allow addresses from other chains to be stored on ENS. Reverse resolution can be expanded to allow the forward resolution to first check the records for the chain that the user is on.
+
+
+## Specification
+
+### Overview
+
+* Reverse registrars will be setup on each L2, with a corresponding registry
+* Reverse registrar will only allow setting the name, without resolver customisability. This is to allow a consistent storage slot that can be checked on L1.
```
Arachnid, line 27: The current version also supports setting text records.

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
Arachnid: At a minimum this needs to be clearer than "if none is found". In general this needs to be much more specific and actionable - describe the specific methods and calls that need to be made and what checks should be done on the return values. When that's already specified elsewhere - eg, forward resolution - reference the relevant standard.

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
Arachnid: Also, the examples should be written descriptively - describing what the client does in the example scenario. SHOULd and MUST etc are for the specification itself.

Arachnid (commented):
This also needs updating to reflect the decision to hex-encode chain IDs.

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
jefflau: I've expanded on this considerably to make it more explicit

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,110 @@
...
+11) If `name` is an empty string, the dapp MUST assume that a primary name does not exist and instead show the address instead. The name resolution process ends here.
+12) If the dapp finds an ENS name, it MUST first check the forward resolution. This can be done by constructing the ENS node for the primary name found: `primaryName = namehash(name)`
+13) It MUST then call the registry to retrieve the resolver `registry.resolver(primaryName)`
+14) Once it has the resolver it MUST call the addr function with the appropriate coinType. If a name was found on step 6, it must use the `coinType` from the chainId of the user. If a name was found at step 9, which is the default Primary name for EoAs, it must check the coinType of chainId 0, which represents [an EoA across all EVM chains](https://namespaces.chainagnostic.org/eip155/caip10). `resolvedAddress = resolver.addr(coinType)`
+15) If `resolvedAddress == address`, the dapp SHOULD consider the Primary Name verified, and can now show this instead of the address within the application.
+16) If `resolverAddress != address` the dapp MUST consider the Primary Name unverified and MUST show the address instead.
```
TateB, line 57:
```suggestion
16) If `resolvedAddress != address` the dapp MUST consider the Primary Name unverified and MUST show the address instead.
```

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,109 @@
...
+* coinType in all instances will be the hex representation to reduce the length of the name
+
+### Deployment and discovery of EVM Reverse registrars
+
+When a new EVM reverse registrar is deployed, it will need to be setup by the owner of the `reverse` node, to setup a subdomain [coinTypeAsHex].reverse. It must then be setup with an Offchain resolver that has an onchain L1 address record that return the contract address of the reverse registrar for that chain. The Offchain resolver will also support wildcard of all the address subdomains with the format [address].[coinTypeAsHex].reverse. Additionally there must be a new EVM gateway setup to handle the CCIP-read revert errors that will go to the corresponding network to gather the chain-specific reverse record and verify this data on L1.
+
+### Resolving Primary ENS Names by a dapp
+
+1) If a dapp has a connected user it SHOULD first check the chainId of the user.
+2) It MUST then construct the cointype using ENSIP11: `coinType = 0x80000000 | chainId`, which must be converted to the hexadecimal representation.
```
Arachnid, line 42:
```suggestion
2) It MUST then construct the cointype using ENSIP11: `coinType = 0x80000000 | chainId`, which must be converted to lower-case hexadecimal representation.
```

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,109 @@
...
+* Discovery of the reverse registrar on each chain will be done by looking up the `addr()` of [coinTypeAsHex].reverse.
+* coinType in all instances will be the hex representation to reduce the length of the name
+
+### Deployment and discovery of EVM Reverse registrars
+
+When a new EVM reverse registrar is deployed, it will need to be setup by the owner of the `reverse` node, to setup a subdomain [coinTypeAsHex].reverse. It must then be setup with an Offchain resolver that has an onchain L1 address record that return the contract address of the reverse registrar for that chain. The Offchain resolver will also support wildcard of all the address subdomains with the format [address].[coinTypeAsHex].reverse. Additionally there must be a new EVM gateway setup to handle the CCIP-read revert errors that will go to the corresponding network to gather the chain-specific reverse record and verify this data on L1.
+
+### Resolving Primary ENS Names by a dapp
+
+1) If a dapp has a connected user it SHOULD first check the chainId of the user.
```
Arachnid, line 41: I'm not sure the use of SHOULD and MUST is appropriate here. It's describing an algorithm, and the client doesn't really have any choice about which bits of the algorithm it does; if it doesn't do all of them as-written, it won't work.

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,109 @@
...
+2) It MUST then construct the cointype using ENSIP11: `coinType = 0x80000000 | chainId`, which must be converted to the hexadecimal representation.
+3) If the chainId is not 1, it SHOULD then construct the ENS node representing the reverse node on that network using the coinType and the user's connected `address`: `node = namehash([address].[coinTypeAsHex].reverse)`
```
Arachnid, line 43: What if the chainId *is* 1?

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,109 @@
...
+4) Call the registry with the namehash to retrieve the resolver `resolver = registry.resolver(node)`
+5) Call name on the resolver `name = resolver.name(node)`
+6) If a string is found, skip to step 12
+7) If `name` is an empty string, it SHOULD check the default primary name for all EVM chains by constructing its node as follows: `node = namehash([address].default.reverse)`.
+8) Call the registry with the namehash to retrieve the resolver `resolver = registry.resolver(node)`
+9) Call name on the resolver `name = resolver.name(node)`
+10) If a string is found, skip to step 12
+11) If `name` is an empty string, the dapp MUST assume that a primary name does not exist and instead show the address instead. The name resolution process ends here.
+12) If the dapp finds an ENS name, it MUST first check the forward resolution. This can be done by constructing the ENS node for the primary name found: `primaryName = namehash(name)`
+13) It MUST then call the registry to retrieve the resolver `registry.resolver(primaryName)`
```
Arachnid, line 53: This implies not using ENSIP-11 wildcard resolution. Instead, this should just specify that it's to use the resolution process documented there.

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,109 @@
...
+14) Once it has the resolver it MUST call the addr function with the appropriate coinType. If a name was found on step 6, it must use the `coinType` from the chainId of the user. If a name was found at step 9, which is the default Primary name for EoAs, it must check the coinType of chainId 0, which represents [an EoA across all EVM chains](https://namespaces.chainagnostic.org/eip155/caip10). `resolvedAddress = resolver.addr(coinType)`
```
Arachnid, line 54: Worth specifying if the legacy `addr` function could/should be used for chainid 1.

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,109 @@
...
+If a resolver is returned for the reverse record, but calling `text` causes a revert or returns an empty string, the client CAN call `.name(namehash('<address>.<coinType>.reverse'))`. If this method returns a valid ENS name, the client MUST:
+
+1. Validate that the reverse record is valid, by resolving the returned name and calling `addr` on the resolver, checking it matches the original evmChainId (converted to cointype) address.
+2. Perform [ENSIP-12 Avatar text record resolution](https://docs.ens.domains/ensip/12) on the primary name.
+
+A failure at any step of this process MUST be treated by the client identically as a failure to find a valid avatar URI.
+
+### Examples of valid L2 reverse resolution
+
+* Arbitrum: 0F32b753aFc8ABad9Ca6fE589F707755f4df2353.2147525809.reverse
```
Arachnid, line 77: These examples still have decimal cointypes.

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,109 @@
...
+
+If we mandated that the address cannot use `addr(node, 60)`, but only the address of the chain in question, it would be possible to use mainnet as a backup. However the fact remains that you would still need to claim and set your Primary ENS name on mainnet, and the possibility for confusion seem to outweigh the benefits of using mainnet (high gas) as a catch-all back up for other L2 EVM chains (low gas). Additionally this would only be useful for EVM-compatible chains and would not benefit non-EVM L2s that have a different address format. 
+
+## Being explicit about default Primary ENS Name
+
+To make things explicit we will require the signing of a message to confirm that the address in question would like to use mainnet or another network as fallback. This would either resolve directly on mainnet. Defaults would only be applicable to EoAs that can sign a message. This is because smart contract accounts would not be able to reliably set a default on all chains.
+
+### Setting default
```
Arachnid, line 100: This is quite vague. Is it worth including at all?

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,109 @@
...
+
+1) Sign a message to set a default record
+2) call `setName()` on the default registrar on L1
+
+Possibility: Allowing L1 Primary ENS names to also use `default.reverse`. This could be incorporated into the public resolver's `name()` function.
```
Arachnid, line 105: ???

Arachnid (changes requested):

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
jefflau: I'm guessing you meant ENSIP-10 here. Will add a reference to wildcard here

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
jefflau: Do you mean removing everything related to defaults?

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,125 @@
...
+
+* Resolver refers to the resolver of the Reverse node whether on L1 or another chain
+
+* Primary Name is the common name use for the reverse node's record `name`
+
+#### Resolution process
+
+1) If a dapp has a connected user it SHOULD first check the chainId of the user.
+2) It will then construct the cointype using ENSIP-11: `coinType = 0x80000000 | chainId`, which must be converted to lower-case hexadecimal representation.
+3) If the chainId is 1, it should call `[address].addr.reverse`, then skip to step 5.
```
Arachnid, line 61: What does it mean to "call" an ENS name?

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,125 @@
...
+13) If the dapp finds an ENS name, it first checks the forward resolution. This can be done by using the resolution processs in [ENSIP-10](https://docs.ens.domains/ensip/10)
+14) In constructing the ENSIP-10 `resolve()` calldata it will construct the addr function with the appropriate coinType. If the chainId was 1, it uses chainId 60 (mainnet) without converting or using the legacy `addr(bytes32 node)` function without an additional cointype argument. If a name was found on step 7, it must use the `coinType` from the chainId of the user. If a name was found at step 10, which is the default Primary name for EoAs, it must check the coinType of chainId 0, which represents [an EoA across all EVM chains](https://namespaces.chainagnostic.org/eip155/caip10). `resolvedAddress = resolver.addr(coinType)`
+15) If `resolvedAddress == address`, the dapp considers the Primary Name verified, and can now show this instead of the address within the application.
+16) If `resolvedAddress != address` the dapp considers the Primary Name unverified and shows the address instead.
```
Arachnid, line 74:
```suggestion
1) Let `chainId` be the chain ID of the DApp's currently connected chain.
2) If `chainId` is 1, set `reverseName = "[address].addr.reverse" and `coinType = 60` and go to step 5.
3) Otherwise, set `coinType` using ENSIP-11: `coinType = 0x80000000 | chainId`.
4) Set `reverseName = '[address].[coinTypeAsHex].reverse'`
5) Set `node = namehash(reverseName)`.
6) Fetch the resolver for the reverse name: `resolver = registry.resolver(node)`
7) Fetch the primary name from the reverse record's resolver: `name = resolver.name(node)`
8) If the primary name is not the empty string, go to step 13.
9) If `name` is an empty string, and `coinType` is not 0, set `reverseName = '[address].default.reverse'` and `coinType = 0` and go to step 5.
12) Otherwise, no primary name exists for this account on this chain; halt and display the address instead.
13) If the dapp finds an ENS name, it MUST then check the forward resolution. This can be done by using the resolution processs in [ENSIP-10](https://docs.ens.domains/ensip/10). When constructing the ENSIP-10 `resolve()` calldata, encode a call to `addr(bytes32 node, uint256 coinType)`. Set `resolvedAddress` to the result of calling `resolve` on the resolver with this calldata.
15) If `resolvedAddress == address`, the dapp considers the Primary Name valid, and can now show this instead of the address within the application.
16) If `resolvedAddress != address` the dapp considers the Primary Name invalid and MUST show the address instead.
```

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,125 @@
...
+    
+Note: The dapp MUST NOT use the reverse record set for Ethereum mainnet ([address].addr.reverse) even if the Primary ENS name has not been set on the target chain, and must treat this identically to an address with no primary name set. 
+
+### Resolving an avatar by a dapp on another EVM chain
+
+ENSIP-12 was concieved before the ENS L2 reverse resolution specification and therefore should be updated to reflect the current state of ENS primary name resolution. This means that all avatar records are able to be updated on a per-chain basis by updating the avatar record on their reverse node.
+
+#### Example of resolving an avatar on an L2 or EVM chain
```
Arachnid, line 82: This is titled "example" but seems to describe a mandatory process for resolving avatars. It also duplicates much of the functionality of the above section. Can we reformulate it to be clearer as a standard, and less duplicative?

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
Arachnid: Yes; it seems incomplete.

Arachnid (changes requested):

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,109 @@
...
+Note: The dapp MUST NOT use the reverse record set for Ethereum mainnet ([address].addr.reverse) even if the Primary ENS name has not been set on the target chain, and must treat this identically to an address with no primary name set. 
+
+### Resolving an avatar by a dapp on another EVM chain
+
+ENSIP-12 was concieved before the ENS L2 reverse resolution specification and therefore should be updated to reflect the current state of ENS primary name resolution. This means that all avatar records are able to be updated on a per-chain basis by updating the avatar record on their reverse node.
+
+To determine the avatar URI for a specific EVM chain address, the client can follow the resolution process above until step 6) And then do the following
+
+1. Perform [Ethereum address avatar text record resolution](https://docs.ens.domains/ensip/12#ethereum-address) on the reverse node.
+2. If none is found, proceed to step 7), if no primary name is found, consider the avatar resolution a failure
```
Arachnid, line 82: Step 7?

ens-improvement-proposals/ensip-17-evmchain-reverse-resolution.md
```diff
@@ -0,0 +1,109 @@
...
+3. Perform [ENS name avatar text record resolution](https://docs.ens.domains/ensip/12#ens-name) on the ENS name.
+4. Complete the Reverse resolution process and consider the avatar found valid if the Primary name has been verified on step 12)
```
Arachnid, line 84: I don't think this last line is necessary.

Arachnid (approved):

docs/ensip/19.mdx
```diff
@@ -0,0 +1,111 @@
...
+To determine the avatar URI for a specific EVM chain address, the client can follow the resolution process above until step 6) And then do the following
+
+1. Perform [Ethereum address avatar text record resolution](https://docs.ens.domains/ensip/12#ethereum-address) on the reverse node.
+2. If none is found, proceed to step 7) in the [Primary name resolution process](#primary-name-resolution-process) above, if no primary name is found, consider the avatar resolution a failure.
+3. Perform [ENS name avatar text record resolution](https://docs.ens.domains/ensip/12#ens-name) on the ENS name.
+
+### Examples of valid L2 reverse resolution
+
+* Arbitrum: 0f32b753afc8abad9ca6fe589f707755f4df2353.8000A4B1.reverse
+* Optimism:
```
cb-elileers, line 91: Unwanted newline?

docs/ensip/19.mdx
```diff
@@ -0,0 +1,111 @@
...
+* Primary Name is the common name use for the reverse node's record `name`
+
+#### Primary Name Resolution process
+
+1) Let `chainId` be the chain ID of the DApp's currently connected chain.
+2) If `chainId` is 1, set `reverseName = "[address].addr.reverse" and `coinType = 60` and go to step 5.
+3) Otherwise, set `coinType` using ENSIP-11: `coinType = 0x80000000 | chainId`.
+4) Set `reverseName = '[address].[coinTypeAsHex].reverse'`
+5) Set `node = namehash(reverseName)`.
+6) Fetch the resolver for the reverse name: `resolver = registry.resolver(node)`
```
stevieraykatz, line 67: I think that this should explicitly state that clients use wildcard resolution so that CCIP-enabled reverse nodes are picked up.

PR #160, opened 2023-08-28
App-specific records

jefflau:
WIP.

Things that need to be thought about:

- [ ] What is the best way to generate unique keys?
- [ ] What's the best use case for this?
- [ ] What's the best way to discover app-specific keys - or does it not matter if only the apps use it?

### DIFF
--- a/ens-improvement-proposals/ensip-xx-app-specific-records.md
+++ b/ens-improvement-proposals/ensip-xx-app-specific-records.md
@@ -1,8 +1,8 @@
 ---
-description: App specific address resolution
+description: App-specific records
 ---
 
-# ENSIP-XX: App specific address resolution
+# ENSIP-XX: App-specific records
 
 | **Author**    | Jeff Lau \<jeff@ens.domains> |
 | ------------- | ---------------------------------- |
@@ -11,7 +11,9 @@ description: App specific address resolution
 
 ### Abstract
 
-This ENSIP extends [ENSIP 9 (multichain address resolution)](ensip-9-multichain-address-resolution.md) and [ENSIP 11 (multichain address resolution)](ensip-11-multichain-address-resolution.md) to include a way to arbitrarily encode an app-specific key to allow app-specific address records. This allows a separation of concerns between a user's main Ethereum/EVM-chain account and a signing key specific to an application allowing easy rotation of those keys.
+This ENSIP extends [ENSIP 9 (multichain address resolution)](ensip-9-multichain-address-resolution.md) and [ENSIP 11 (multichain address resolution)](ensip-11-multichain-address-resolution.md) and [ENSIP 5 (Text Records)](ensip-5-text-records.md) to include a way to app specific addresses and public keys.
+
+This allows a separation of concerns between a user's main Ethereum/EVM-chain account and a signing key specific to an application allowing easy rotation of those keys. 
 
 ### Motivation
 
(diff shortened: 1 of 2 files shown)


ens-improvement-proposals/ensip-xx-app-specific-records.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Abstract
+
+This ENSIP extends [ENSIP 9 (multichain address resolution)](ensip-9-multichain-address-resolution.md) and [ENSIP 11 (multichain address resolution)](ensip-11-multichain-address-resolution.md) and [ENSIP 5 (Text Records)](ensip-5-text-records.md) to include a way to app specific addresses and public keys.
+
+This allows a separation of concerns between a user's main Ethereum/EVM-chain account and a signing key specific to an application allowing easy rotation of those keys. 
+
+### Motivation
+
+Apps such as Farcaster that use Sign in with Ethereum as a primary login require the generation of a new EoA. It is also possible that you may want to split your address usage across different applications for address hygiene, security or privacy reason. Lastly with account abstraction becoming more popular, we may end up with an increasing amount of different addresses for different applications that man need to be connected.
```
varunsrin, line 20: A technicality, but Farcaster wallet clients don't use the SIWE standard for account creation / usage.  A new account is created when you make a transaction to a contract, and then the address that made the transaction is now the owner of the account.

ens-improvement-proposals/ensip-xx-app-specific-records.md
```diff
@@ -0,0 +1,76 @@
...
+
+A new resolver interface is defined, consisting of the following method:
+
+```solidity
+interface IBinaryResolver {
+  /// @notice Returns the binary data associated with a key for an ENS name
+  /// @param node A nodehash for an ENS name
+  /// @param key A key to lookup text data for
+  /// @return The binary data
+  function binaryRecord(bytes32 node, string key) view returns (bytes data);
```
varunsrin, line 47: Is there a standard for what binary data should be returned for a record? Or is that up to each app to decide?

ens-improvement-proposals/ensip-xx-app-specific-records.md
```diff
@@ -0,0 +1,76 @@
...
+### Example use case
+
+* If you have address 0x123... on Optimism
+* You setup a reverse record for 0x123... to example.eth
+* You setup forward resolution for optimism to 0x123 to complete reverse resolution
+* You create 0x456... (could be non-ethereum address) on Farcaster
+* You set the farcaster record on example.eth to 0x456...
+* You now have a connection from 0x123... (your main optimism address) to 0x456... your farcaster address
+* You add your ENS name to your account settings in Farcaster (stored in FC's hub servers)
+* Now Farcaster can check your farcaster address in example.eth and show records from example.eth or NFTs you own on example.eth
```
varunsrin, line 60: you might want this to be double opt-in. otherwise i could register something like poop.eth and point it at your accounts keys without your permission. you could require that the binary data in the record contain a signature from the target address

ens-improvement-proposals/ensip-xx-app-specific-records.md
```diff
@@ -0,0 +1,76 @@
...
+* You can take the private key and put it onto any device, and login to Farcaster seemlessly by putting the key on the device
+* This key gives access to your Farcaster account, but if a device is compromised, you can rotate the key to a new account rather than having to transfer all your assets out of your main account.
```
varunsrin, line 62: how would this part work? if you are concerned your keys are compromised, and you own some eth or nfts there, youd still have to manually transfer them to the new account

ens-improvement-proposals/ensip-xx-app-specific-records.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Considerations
+
+Should the spec also deal with chains? E.g.
```
varunsrin, line 66: more broadly, apps should be able to specify multiple keys. for example, farcaster may use different keys for onchain actions vs encrypted messaging. so if you have a general format that apps can register subkeys with, they can easily extend them with chain info if needed.

varunsrin:
Overall this takes the same approach that Farcaster took with the [KeyRegistry](https://github.com/farcasterxyz/contracts/blob/main/src/KeyRegistry.sol) but with a few differences: 

1. No validation of keys 
2. No attribution to key requestor 
3. No explicit double opt-in 

I think 1 & 2 are fine and a general purpose system like ENS probably shouldn't have these specifics that may not apply to all apps. But i think 3 is worth considering as part of the spec

gskril:
Cleaning up — Jeff confirmed this is safe to close. If we want to revive the idea, it should be reopened in the [canonical ENSIPs repo](https://github.com/ensdomains/ensips)

PR #161, opened 2023-09-25
Canonical name resolver

jefflau:
Adds a standard for a resolver that allows DNS CNAME-like functionality to resolvers.

### DIFF
--- a/ens-improvement-proposals/ensip-xx-canonical-name-resolver.md
+++ b/ens-improvement-proposals/ensip-xx-canonical-name-resolver.md
@@ -1,0 +1,51 @@
+---
+description: Allows names to set a cname to alias records to another name
+---
+
+# ENSIP-XX: CNAME resolver
+
+| **Author**    | Jeff Lau \<jeff@ens.domains>
+| ------------- | ---------------------------------------------------------------- |
+| **Status**    | Draft                                                            |
+| **Submitted** | 2023-09-25                                                       |
+
+### Abstract
+
+Canonical Name records have existed for DNS since the 1987 in [RFC 1034](https://datatracker.ietf.org/doc/html/rfc1034). They allow aliasing a name to another name e.g. vitalik.eth to records.vitalik.eth. This specification outlines a resolver standard to implement CNAMEs at a resolver level without having to change the base ENS name resolution protocol.
+
+### Specification
+
+If a cname is found on a name, all records that are unset will be expected to proxy through the specified cname's records. Records that are already set on the resolver will ignore the cname.
+
+### Implementation
+
+```solidity
+interface ICnameResolver {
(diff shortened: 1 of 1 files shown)


ens-improvement-proposals/ensip-xx-canonical-name-resolver.md
```diff
@@ -0,0 +1,51 @@
...
+| **Status**    | Draft                                                            |
+| **Submitted** | 2023-09-25                                                       |
+
+### Abstract
+
+Canonical Name records have existed for DNS since the 1987 in [RFC 1034](https://datatracker.ietf.org/doc/html/rfc1034). They allow aliasing a name to another name e.g. vitalik.eth to records.vitalik.eth. This specification outlines a resolver standard to implement CNAMEs at a resolver level without having to change the base ENS name resolution protocol.
+
+### Specification
+
+If a cname is found on a name, all records that are unset will be expected to proxy through the specified cname's records. Records that are already set on the resolver will ignore the cname.
```
Arachnid, line 18: This should clarify that it's the resolver's reponsibility to implement CNAME functionality, not the client.

ens-improvement-proposals/ensip-xx-canonical-name-resolver.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Implementation
+
+```solidity
+interface ICnameResolver {
+    event CNameChanged(bytes32 indexed node, string name);
+    function cname(bytes32 node) external view returns (string memory);
```
Arachnid, line 25: I think we need to make a decision here: Either cname takes a node and doesn't implement CCIP-Read and wildcards (so only onchain names can be CNAMEd to), or it takes a DNS-encoded name and does implement CCIP-Read and wildcards. The latter is the more future-proof option, but this implementation requirement needs to be called out.

ens-improvement-proposals/ensip-xx-canonical-name-resolver.md
```diff
@@ -0,0 +1,51 @@
...
+}
+```
+
+```solidity
+contract CnameResolver {
+    function addr(bytes32 node, uint256 coinType){
+        bytes memory currentAddr = versionable_addresses[recordVersions[node]][node][coinType];
+
+        if(currentAddr.length == 0 && cname.length != 0){
```
Arachnid, line 34: This will have to be a lot more complex to support CCIP Read and wildcards.

Perhaps we should just implement a canonical CNAME resolver mixin instead of writing a spec?

Arachnid (changes requested):

gskril:
Cleaning up — Jeff confirmed this is safe to close. If we want to revive the idea, it should be reopened in the [canonical ENSIPs repo](https://github.com/ensdomains/ensips)

PR #165, opened 2023-10-31
[Proposal] ENSIP-17: DataURI Format in Contenthash

sshmatrix:
This proposed extension to ENSIP-07 introduces `data:uri` format in ENS Contenthash field, allowing dynamic data streaming using EIP-3668: CCIP-Read and ENSIP-10: Wildcard Resolution. Please feel free to go through the draft and ask any questions, seek clarifications, give suggestions or propose edits.

### DIFF
--- a/ens-improvement-proposals/ensip-17-datauri-contenthash.md
+++ b/ens-improvement-proposals/ensip-17-datauri-contenthash.md
@@ -1,6 +1,6 @@
-**DESCRIPTION:** DataURI (RFC-2397) Format in Contenthash
+# ENSIP-17: DataURI Format in Contenthash
 
-# ENSIP-XX: DataURI Format in Contenthash
+### RFC-2397 Compliant DataURI Format in Contenthash
 
 | &nbsp; | &nbsp; |
 | ------ | ------ |
@@ -14,11 +14,11 @@ This ENSIP introduces DataURI format in Contenthash field (ENSIP-07) for compati
 
 ## Motivation
 
-ENS `contenthash` (ENSIP-07) currently enables linking to static content which is strictly off-chain. The off-chain content is entirely dependent on off-chain providers, and updating this content for ENS-based decentralised websites typically requires updating the on-chain `contenthash` explicitly (except for IPNS). ENS domains' `avatar` text records and their ERC-721/-1155 interfaces already support generated DataURI bytes (`data:uri`) to resolve JSON and image metadata. This specification enables a similar `data:uri` format in ENS `contenthash` field, allowing ENS Resolvers to fetch and serve on-chain and/or off-chain data. The off-chain resources for the DataURI content may use CCIP-Read and an appropriate `utf-8` decoder to render the encoded bytes.
+ENS `contenthash` (ENSIP-07) currently enables linking to static content which is strictly off-chain. The off-chain content is entirely dependent on off-chain providers, and updating this content for ENS-based decentralised websites typically requires updating the on-chain `contenthash` explicitly (except for IPNS). ENS domains' `avatar` text records and their ERC-721/-1155 interfaces already support generated DataURI bytes (`data:uri`) to resolve JSON and image metadata. This specification enables a similar `data:uri` format in ENS `contenthash` field, allowing ENS Resolvers to fetch and serve on-chain and/or off-chain data. The off-chain resources for the DataURI content may use CCIP-Read and an appropriate `utf-8` decoder to render the encoded bytes. This specification allows complete support for dynamic data in ENS Contenthash using CCIP-Read ([EIP-3886](https://eips.ethereum.org/EIPS/eip-3668)) and Wildcard Resolution ([ENSIP-10](https://docs.ens.domains/ens-improvement-proposals/ensip-10-wildcard-resolution)).
 
 ## Specification
 
-This specification is an extension of [ENSIP-07](https://docs.ens.domains/ens-improvement-proposals/ensip-7-contenthash-field) to support in-line bytes of data conforming to the `data:uri` scheme (RFC-2397) as ENS Contenthash. There are no changes to be made in the current ENS Resolvers since `contenthash` bytes are parsed as `utf-8` characters by default. Only a standardisation needs to be enacted for web3 providers to begin resolving ENS Contenthash in `data:uri` scheme. Simple details of the proposed standardisation are as follows:
+This specification is an extension of [ENSIP-07](https://docs.ens.domains/ens-improvement-proposals/ensip-7-contenthash-field) to support in-line bytes of data conforming to the `data:uri` scheme ([RFC-2397](https://datatracker.ietf.org/doc/html/rfc2397)) as ENS Contenthash. There are no changes to be made in the current ENS Resolvers since `contenthash` bytes are parsed as `utf-8` characters by default. Only a standardisation needs to be enacted for web3 providers to begin resolving ENS Contenthash in `data:uri` scheme. Simple details of the proposed standardisation are as follows:
 
 ### Decoded String
 
@@ -51,8 +51,18 @@ With this simple standardisation, web3 providers may now serve `data:uri` conten
 
 ### Implementation
 
-**GitHub** : https://github.com/namesys-eth/datauri-eth-resolver (`Work-In-Progress`)
+**GitHub** : [`namesys-eth/datauri-eth-resolver`](https://github.com/namesys-eth/datauri-eth-resolver) (`Work-In-Progress`)
+
+### References
+
+[1] `ENSIP-07`: [Contenthash Field](https://docs.ens.domains/ens-improvement-proposals/ensip-7-contenthash-field)
+
+[2] `ENSIP-10`: [Wildcard Resolution](https://docs.ens.domains/ens-improvement-proposals/ensip-10-wildcard-resolution)
+
+[3] `EIP-3668`: [CCIP Read: Secure Off-Chain Data Retrieval](https://eips.ethereum.org/EIPS/eip-3668)
+
+[4] `RFC-2397`: [The "data" URL Scheme](https://datatracker.ietf.org/doc/html/rfc2397)
 
 ### Copyright
 
-Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
+Copyright and related rights waived via [`CC0`](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/ens-improvement-proposals/ensip-xx-datauri-contenthash.md
+++ b/ens-improvement-proposals/ensip-xx-datauri-contenthash.md
@@ -1,0 +1,58 @@
+**DESCRIPTION:** DataURI (RFC-2397) Format in Contenthash
+
+# ENSIP-XX: DataURI Format in Contenthash
+
+| &nbsp; | &nbsp; |
+| ------ | ------ |
+| **Author(s)** | [sshmatrix.eth](@sshmatrix), [ethlimo.eth](@ethlimo), [freetib.eth](@0xc0de4c0ffee) |
+| **Status**    | Draft |
+| **Submitted** | `2023-10-31` |
+
+### Abstract
+
+This ENSIP introduces DataURI format in Contenthash field (ENSIP-07) for compatible ENS resolvers. DataURI format (RFC-2397) is desired and suitable for enabling dynamic dWeb content for ENS domains using on-chain and/or off-chain resources. 
+
+## Motivation
+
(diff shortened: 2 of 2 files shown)


Arachnid:
Thanks for submitting this! It's nice to see expanded standardisation of this, and I like the novel combination of CCIP read and contenthash.

However, contenthash is a binary field, as defined here: https://docs.ens.domains/ens-improvement-proposals/ensip-7-contenthash-field

> The value returned by contenthash MUST be represented as a machine-readable [multicodec](https://github.com/multiformats/multicodec).

This spec will need to be updated to either use an existing multicodec type to encode the data, or to use a newly defined one.

sshmatrix:
> This spec will need to be updated to either use an existing multicodec type to encode the data, or to use a newly defined one.

We were looking to use [`plaintextv2`](https://github.com/multiformats/multicodec/blob/3122fd93aecf1b765df6d9bf2641cbe80f4619b2/table.csv#L522) as `hex("pla") = 0x706c61` from multicodec. 
[522](https://github.com/multiformats/multicodec/blob/3122fd93aecf1b765df6d9bf2641cbe80f4619b2/table.csv#L522)| plaintextv2 | multiaddr | 0x706c61 | draft
-- | -- | -- | -- | --

So we're deviating from multicodec on this to implement a [default/fallback profile](https://github.com/ensdomains/content-hash/blob/e3f00395eb8ca0b7b6422e611a363a294629ecb8/src/profiles.ts#L186
) implemented in some JS code e.g. `ens/contenthash.js`

```
 default: {
    encode: encodes.utf8,
    decode: decodes.utf8,
  },
```
We can request to add `stringTohex("data:")` = `0x646174613a` in multicodec too. But we prefer reuse NFT DataURIs directly without any extra prefix for namespace, version & multiaddr/version.
For instance, contenthash for `1234.hello-nft.eth` can resolve `bytes(tokenURI(1234))` directly as `data:application/json,{...metadata}`.

Arachnid:
> But we're eventually deviating from multicodec on this to feature a [default/fallback profile](https://github.com/ensdomains/content-hash/blob/e3f00395eb8ca0b7b6422e611a363a294629ecb8/src/profiles.ts#L186) implemented in some JS code e.g. `ens/contenthash.js`

The problem with this is that it directly contradicts ENSIP-7, as quoted above. Contenthashes *must* be valid multicodec values. We're not going to standardise something that contradicts an earlier standard like this.

sshmatrix:
This is a copy of the [response](https://discuss.ens.domains/t/draft-ensip-17-datauri-format-in-contenthash/18048/7) posted in the ENS Forum.

---

Thanks for the feedback! We have looked into possible ways for this draft ENSIP to be compatible with multicodec. These are our findings in form of different implementations with and without multicodec. We are open to either implementation in the end and update this draft ENSIP as required.

## A) Bypass Multicodec:

First, we'd like to point to the current state of `ens/content-hash.js`. When using `hex("data:") = 0x646174613a` as prefix, encoding doesn't work and the decoder nearly works but it removes the first byte in the process. Please see example below,

```js
import {encode, decode } from "@ensdomains/content-hash";
console.log(encode("data:text/plain;base64,SGVsbG8gV29ybGQ"))
//> 00000000000000000000

console.log(decode("646174613a746578742f706c61696e3b6261736536342c534756736247386756323979624751"))
//> ata:text/plain;base64,SGVsbG8gV29ybGQ
```

The extra `0x00` prefix identifier as a spacer/pseudo namespace could prevent any future collision with multicodec.

```js
console.log(decode("00646174613a746578742f706c61696e3b6261736536342c534756736247386756323979624751"))
//> data:text/plain;base64,SGVsbG8gV29ybGQ
```

Quoting @raffy here,
[quote="raffy, post:3, topic:18048"]
Although for simplicity, I’m a fan of bypassing the multicodec stuff (as long as the first uvarint decodes correctly) and just embedding raw utf-8 data.
[/quote]

This will bypass multicodec for DataURIs in Contenthash. `ens/content-hash.js` and any gateways/clients can easily implement this with basic checks, i.e. checking for *if* prefix is `0x00646174613a` before encoding and decoding so that ENS clients and gateways can use DataURIs directly without leaving any room for current or future collisions with multicodec formats. This approach will be ENS specific and we can change our ENSIP draft to reflect this. 

✅ This is our preferred implementation but we are not married to it.

## B) Multicodec Compatible Formats

If multicodec must be used, then we'd like to propose the following options:
 
### 1) `raw` data type with IPFS `namespace`:

IPFS `namespace` is compatible with DataURIs using `raw` data type.

```js
import { CID } from 'multiformats/cid'
import { identity } from 'multiformats/hashes/identity'
import * as raw from "multiformats/codecs/raw";
const utf8 = new TextEncoder();

let data = utf8.encode('data:text/plain;base64,SGVsbG8gV29ybGQ')
let cid = CID.create(1, raw.code, await identity.digest(data))
```
#### IPFS Format :
```
base32: bafkqajtemf2gcotumv4hil3qnrqws3r3mjqxgzjwgqwfgr2wonreoodhkyzds6lci5iq
base16: f01550026646174613a746578742f706c61696e3b6261736536342c534756736247386756323979624751
Contenthash: 0xe30101550026646174613a746578742f706c61696e3b6261736536342c534756736247386756323979624751
```

|Namespace|Length|Version|Multicodec|Multihash|Length|Data|
|---|---|---|---|---|---|---|
|`ipfs`|`1`|`1`|`raw`|`identity`|`38`|`data:text/plain;base64,SGVsbG8gV29ybGQ`|
|`0xe3`|`0x01`|`0x01`|`0x55`|`0x00`|`0x26`|`0x646174613a746578742f706c61696e3b6261736536342c534756736247386756323979624751`|

> CID Inspector : [https://cid.ipfs.tech/#bafkqajtemf2gcotumv4hil3qnrqws3r3mjqxgzjwgqwfgr2wonreoodhkyzds6lci5iq](https://cid.ipfs.tech/#bafkqajtemf2gcotumv4hil3qnrqws3r3mjqxgzjwgqwfgr2wonreoodhkyzds6lci5iq) 

> Public Gateway : https://ipfs.io/ipfs/bafkqajtemf2gcotumv4hil3qnrqws3r3mjqxgzjwgqwfgr2wonreoodhkyzds6lci5iq

Since this method uses IPFS namespace, `ens/content-hash.js` and any compatible gateway or client must check if the encoded payload is using `raw` as multicodec with `identity` (`blank`) as multihash; the shorthand prefix for this is `0xe301015500`. Clients can decode the remaining raw data as `utf-8` string; if this data is not `data:uri` formatted, it should be auto-rendered as `plaintext` for correctly formatted `data:uri` clients, and gateways can render according to mime or type included in the DataURI payload.

### 2) `plaintextv2` data with IPFS or IPLD `namespace`:

This is similar to the previous option but using `plaintextv2` instead of `raw` as multicodec. 

[quote="0xc0de4c0ffee, post:4, topic:18048"]
There’s [`plaintextv2`](https://github.com/multiformats/multicodec/blob/3122fd93aecf1b765df6d9bf2641cbe80f4619b2/table.csv#L522) as `hex("pla") = 0x706c61` multiaddr prefix in multicodec.

[522](https://github.com/multiformats/multicodec/blob/3122fd93aecf1b765df6d9bf2641cbe80f4619b2/table.csv#L522) plaintextv2 multiaddr 0x706c61 draft
[/quote]

```js
import { CID } from 'multiformats/cid'
import { identity } from 'multiformats/hashes/identity'
const utf8 = new TextEncoder();

let data = utf8.encode('data:text/plain;base64,SGVsbG8gV29ybGQ');
let cid = CID.create(1, 0x706c61, await identity.digest(data))
```

#### `plaintextv2` format using IPFS namespace

```
Base32 : bahq5rqidaatgiylume5hizlyoqxxa3dbnfxdwytbonstmnbmkndvm43ci44govrshf4wer2r
Base16 : f01e1d8c1030026646174613a746578742f706c61696e3b6261736536342c534756736247386756323979624751
Contenthash : 
0xe30101e1d8c1030026646174613a746578742f706c61696e3b6261736536342c534756736247386756323979624751
```

|Namespace|Length|Version|Multicodec|Length|Multihash|Length|Data|
|---|---|---|---|---|---|---|---|
|`ipfs`|`1`|`1`|`plaintextv2`|3|`identity`|`38`|`data:text/plain;base64,SGVsbG8gV29ybGQ`|
|`0xe3`|`0x01`|`0x01`|`0xe1d8c1`| `0x03`|`0x00`|`0x26`|`0x646174613a746578742f706c61696e3b6261736536342c534756736247386756323979624751`|

> CID Inspector : [https://cid.ipfs.tech/#bahq5rqidaatgiylume5hizlyoqxxa3dbnfxdwytbonstmnbmkndvm43ci44govrshf4wer2r](https://cid.ipfs.tech/#bahq5rqidaatgiylume5hizlyoqxxa3dbnfxdwytbonstmnbmkndvm43ci44govrshf4wer2r) 

> Public Gateway : https://ipfs.io/ipfs/bahq5rqidaatgiylume5hizlyoqxxa3dbnfxdwytbonstmnbmkndvm43ci44govrshf4wer2rh

#### `plaintextv2` format using IPLD namespace

```
Base32 : bahq5rqidaatgiylume5hizlyoqxxa3dbnfxdwytbonstmnbmkndvm43ci44govrshf4wer2r
Base16 : f01e1d8c1030026646174613a746578742f706c61696e3b6261736536342c534756736247386756323979624751
Contenthash : 
0xe20101e1d8c1030026646174613a746578742f706c61696e3b6261736536342c534756736247386756323979624751
```

|Namespace|Length|Version|Multicodec|Length|Multihash|Length|Data|
|---|---|---|---|---|---|---|---|
|`ipld`|`1`|`1`|`plaintextv2`|3|`identity`|`38`|`data:text/plain;base64,SGVsbG8gV29ybGQ`|
|**`0xe2`**|`0x01`|`0x01`|`0xe1d8c1`| `0x03`|`0x00`|`0x26`|`0x646174613a746578742f706c61696e3b6261736536342c534756736247386756323979624751`|

> CID Inspector : [https://cid.ipfs.tech/#bahq5rqidaatgiylume5hizlyoqxxa3dbnfxdwytbonstmnbmkndvm43ci44govrshf4wer2r](https://cid.ipfs.tech/#bahq5rqidaatgiylume5hizlyoqxxa3dbnfxdwytbonstmnbmkndvm43ci44govrshf4wer2r) 

> Public Gateway : https://dweb.link/api/v0/dag/get?arg=bahq5rqidaatgiylume5hizlyoqxxa3dbnfxdwytbonstmnbmkndvm43ci44govrshf4wer2r

❗ **NOTE:** `plaintextv2` is still in draft and IPFS gateways CANNOT yet render it properly resulting in 500 error. Trying to use it as IPLD might require changing the encoding process too. We do not prefer this method.

## C) `CARv1` strings

`CARv1` files as strings can represent IPFS data or IPLD files and directory but this implementation is more complex than previous options so we'll only mention this as a footnote. We don't have the bandwidth to implement this. If ENS devs are happy to explore this for future implementation, it'll be one of best options for fully on- or off-chain generators and IPFS data storage.

Based on this, we are happy to get more feedback and then make changes to the draft ENSIP!

Arachnid:
I'm confused why IPFS is involved at all. Why can't you either use an existing multicodec identifier or define a new one for URIs?

sshmatrix:
> I'm confused why IPFS is involved at all. Why can't you either use an existing multicodec identifier or define a new one for URIs?

We had thought about the option of a new namespace in our draft but we skipped it since IPFS/IPLD namespace with plaintext or raw encoded payload contained within the CID is sufficiently unique, and backwards compatible with IPFS gateways returning plaintext data. 

```
Proposed IPFS (raw-ipld/plaintext) : 0xe301015500 + <data.length> + <data>
  Normal IPFS      (dag-pb/sha256) : 0xe301017012 + <hash.length> + <hash of data or dag>
```
**However**, we're open to requesting a new ENS-specific namespace for this ENSIP only. In this regard, please suggest a short code (>=2 bytes) for this and we'll PR that in the multicodec table soon. Something like

- `hex('ens') = 0x656e73` sounds like a good option to us; equivalent namespace is VARINT(`0x656e73`) = `0xf3dc9503`
- Non-ASCII option `0xda7a` is also good, which will lead to a VARINT(`0xda7a`) = `0xfab403` namespace

The above two options with `raw` multiaddr should implement like:

|&nbsp;|Namespace|Version|Multiaddr|Multihash|Length|Example|
|---|---|---|---|---|---|---|
|&nbsp;|&nbsp;|`1`|`raw`|`identity`|`38`|`data:text/plain;base64,SGVsbG8gV29ybGQ`|
|`0xda7a`|`0xfab403`|`0x01`|`0x55`|`0x00`|`0x26`|`0x646...751`|
|`hex('ens')`|`0xf3dc9503`|`0x01`|`0x55`|`0x00`|`0x26`|`0x646...751`|

There are currently no DataURI-related multiaddr or namespaces and we do not want to introduce one in this context due to lack of manpower and funding to follow up on sidequests. DataURI class is too broad and it'll also require mime/type codecs which are pending on issues or PR for a very long time. See below:

> https://github.com/multiformats/multicodec/pull/159 

> https://github.com/multiformats/multicodec/pull/4

PR #194, opened 2024-03-05
Simplify ENSIP 16

makoto:
This changes simplify the `metadata` function to only return `graphql` url. All the rest of the information should be fetched via [EIP 5559](https://eips.ethereum.org/EIPS/eip-5559)

The Reference implementation is at  https://github.com/ensdomains/evmgateway/pull/34/files


lucemans (approved):
LGTM

Arachnid (approved):

PR #204, opened 2024-03-08
Updating Stewards and Token recovery information

alexvansande:
Updating the stewards page to include the 2024 update.

Also decided to add the token recovery process to the docs so that it's more findable by users.


lucemans:
Looking good so far! Will update links and merge!

lucemans (approved):
LGTM

PR #221, opened 2024-04-03
Patches for rich data

Jontes-Tech:
[x] Add Opengraph image to JSONLD
[x] Remove breadcrumbs without a name (which are created for some reason)
[x] Use absolute URLs for breadcrumbs (as per the spec)


lucemans (approved):
LGTM

PR #245, opened 2024-05-21
[DRAFT] ENSIP Verifiable Credentials & Presentations

lucemans:


gskril:
Maybe add the fact that some names use JSON-stringified array instead of other types under backwards compatibility

gskril:
Moved to https://github.com/ensdomains/ensips/pull/3

PR #254, opened 2024-06-11
Update Navigation, Fallback, Code Snippets, Reverse Resolution, and I…

lucemans:
…mages

image
image
image

### DIFF
--- a/docs/web/index.mdx
+++ b/docs/web/index.mdx
@@ -10,6 +10,60 @@ export const meta = {
 
 This section walks you through how to leverage the ENS open standards to improve the user experience of your app.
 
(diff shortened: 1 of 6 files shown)


docs/web/index.mdx
gskril, line 1: Should some of these new elements be broken out into reusable components?

gskril, line 1: Fallback image doesn't seem to be working for luc.cash and puma.eth. I assume this is because `https://docs.ens.domains/fallback.svg` won't exist until this PR is merged but wanted to double check

image

lucemans: TODO: update to dark border color

gskril (approved):

PR #269, opened 2024-08-08
Expanding contract interfaces template, filled out for resolver page

serenae-fansubs:
Generic interface templating moved to `interfaces.tsx`. Each contract/interface page just needs to call `interfaceDetails`.

Added `ContractParameter`, where both inputs and outputs can be specified. Each parameter/result can have its own specific description.

The `seeMore` field will automatically have EIP/ENSIP keys turned into links.

A new `examples` field can be used to show usage examples for methods.

And other small tweaks.


serenae-fansubs:
@lucemans The build check is failing, but I think that's due to something unrelated to this PR:

```
Error: Static page generation for /dns/tlds is still timing out after 3 attempts.
```

lucemans:
Rebased, going to test CI

lucemans (approved):
LGTM

PR #271, opened 2024-08-12
fet-1490: basic CSP

staslysak:


lucemans:
Closing this as we do not use server-side rendering and thus no next middlewares

PR #283, opened 2024-09-05
Adding UR page

serenae-fansubs:


docs/resolvers/universalresolver.mdx
```diff
@@ -0,0 +1,58 @@
...
+        'luc.eth',
+        'serenae.eth'
+    ]
+};
+
+# Universal Resolver
+
+## Overview
+
+The Universal Resolver is a contract that handles the work of resolving a name entirely onchain, making it possible to make a single smart contract call to resolve an ENS name.
```
gskril, line 20: Probably worth mentioning that the UR is used under the hood in viem and ensjs, and devs don't usually have to interact with it directly. This doc is for advanced use.

gskril (approved):
Left 1 suggestion but otherwise lgtm!

docs/resolvers/universalresolver.mdx
serenae-fansubs: Done!

PR #289, opened 2024-09-16
Remove broken link

LeonmanRolls:


gskril:
Asking metagov if there's an updated link we should use instead

lucemans (approved):

estmcmxci:
I'm opening a PR to add the updated link.

PR #331, opened 2024-10-14
Update CCIP-Read POST example url: omit `{sender}` and `.json`

fselmo:
For clarity, the POST url doesn't seem to necessitate a `{sender}` and it also doesn't seem to make sense to POST to a `.json`. When we implemented CCIP-Read in web3.py, there was a bit of ambiguity on whether we should still include a check for `{sender}` in the url based on this example ([PR for this change](https://github.com/ethereum/web3.py/pull/3291)). I think this change would help clarify this for readers / implementers.

### DIFF
--- a/docs/resolvers/ccip-read.mdx
+++ b/docs/resolvers/ccip-read.mdx
@@ -92,14 +92,15 @@ The URL for this gateway is determined by the `OffchainLookup` error, and is pas
 
 ### Implement Gateway Endpoint {{ navtitle: 'Implement Gateway', id: 'endpoint' }}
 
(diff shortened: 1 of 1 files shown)


Arachnid:
It doesn't necessitate a `{sender}`, but it is *allowed*. I think I left it in here to make that clear.

If you'd like to amend it, can you add a note to the effect that it's optional?

fselmo:
welp... looks like this [was resolved](https://github.com/ensdomains/docs/pull/409/files#diff-3d4d0c65912a7aa738d9448bf491a24cecddf155c923b451806737f80deea67eL109)

PR #341, opened 2024-10-30
Introducing EP 5.18-21 into governance docs

estmcmxci:


gskril (approved):
Thanks for contributing! For the future - would be great if you could also add proposals to the navigation in `protocol.ts` as I did here.

estmcmxci:
Roger that.

PR #345, opened 2024-11-06
Typo corrections

mdqst:
### Typo corrections

* dweb/intro.mdx:

"atleast" → should be written as "at least" (two separate words, not one).

Corrected.

* learn/dns.mdx

"so much more then" → should be "so much more than"
Explanation: The correct expression is "more than," not "more then."

Corrected.

* web/avatars.mdx

"for for" — There is a repeated word. Corrected.

* docs/terminology.mdx

"Cross Chain Interoperability Procol Read" — There is a typo in the term "Procol." It should be "Protocol". The corrected phrase is "Cross Chain Interoperability Protocol Read".

Corrected.


gskril (approved):
Lgtm, thanks!

PR #354, opened 2024-11-14
Adding L2 notes to Naming Contracts page.

serenae-fansubs:
Also adding note to subname registrar page.

### DIFF
--- a/docs/web/naming-contracts.mdx
+++ b/docs/web/naming-contracts.mdx
@@ -102,6 +102,8 @@ From your contract you can execute `setName` function on the [Reverse Registrar]
 
 ## L2 Contracts
 
+Primary Names are not yet supported on L2s. In the meantime, the following sections only apply to testnet ENS deployments. Mainnet support should be coming by the end of January 2025.
+
 ### L2 - Ownable
 
 If you want to set a primary name for a contract you are deploying on an L2 chain, you need to make sure your contract implements the [Ownable](https://docs.openzeppelin.com/contracts/5.x/api/access#Ownable) interface from OpenZeppelin, and has an account you own set as the owner.


gskril:
Instead of relying on `Ownable`, is it possible to just call `setName()` within the constructor directly? I think this is a preferred approach if possible, or at least worth mentioning both options.

serenae-fansubs:
> Instead of relying on `Ownable`, is it possible to just call `setName()` within the constructor directly? I think this is a preferred approach if possible, or at least worth mentioning both options.

I intentionally left that out, because don't think that would be a preferred option, it's something that should be discouraged actually, in my opinion.

First of all, you typically don't know what a contract address is going to be until you deploy it (you can probably simulate and derive that based on nonce, maybe, depending on the chain). So you don't even know what to set your ENS name's ETH record to until _after_ you deploy usually. There is no forward-resolution check on the reverse registrar though, so technically you can set the reverse record first, and then work backwards to set that ETH address on whatever ENS name you set it to. Technically possible, but possibly confusing.

However the bigger issue is that if you just do it in the constructor, then it's a one-time-only thing. You cannot change it from then on. If your ENS name gets compromised, or if you need to migrate to a different ENS name / subname, then you're stuck, because you cannot update the reverse record for that contract anymore. That's why (in my opinion) we don't present that as an option for L1 contracts either.

gskril:
Hmmm I'm thinking along the lines of the "default names" concept — seems like something would have to happen in the constructor for that to work ?

gskril:
Worth mentioning that you can call `setName` in constructor, but its immutable and not necessarily recommended.

Also worth mentioning that the name must forward resolve to the relevant address with the respective cointype for it to actually work.

gskril:
Also would be nice to reference [this list of deployments](https://docs.ens.domains/web/reverse) vs duplicate it.

docs/web/naming-contracts.mdx
```diff
@@ -82,3 +99,23 @@ You can do so by hitting the `setNameForAddr` function on the [Reverse Registrar
 From your contract you can execute `setName` function on the [Reverse Registrar](/registry/reverse)
 
 <LiveDemo id="reversesetnamefordemo" />
+
+## L2 Contracts
+
```
gskril, line 104:
```suggestion

Primary Names are not yet supported on L2s. In the meantime, the following sections only apply to testnet ENS deployments. Mainnet support should be coming by the end of January 2025.

```

gskril:
Idk why the Cloudflare build is failing when the `Build` test is succeeding, so I'm going to merge anyways and can fix that issue separately if it persists.

PR #357, opened 2024-11-17
Misc. fixes and improvements to "Resolvers" section

0x-r4bbit:
There are a few typos and a quoted sentence that doesn't really makes sense where it currently is.

I don't know what the original intention was. I doesn't look like it adds any value so I've removed it instead of leaving it and have readers being confused.

### DIFF
--- a/app/local/data/resolver.tsx
+++ b/app/local/data/resolver.tsx
@@ -2,24 +2,6 @@
 import { CodeGroup } from '../content/prose/code/group/CodeGroup';
 import { ContractMethod } from './interfaces';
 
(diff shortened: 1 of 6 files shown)


gskril:
Thanks for pointing this out and contributing @0x-r4bbit. I'll make some additional changes to add clarity before merging.

0x-r4bbit:
@gskril thanks for the heads-up! Also don't hesitate to tell me if you want anything changed :)

serenae-fansubs (approved):
lgtm!

PR #360, opened 2024-11-20
Add clarity about resolution starting from L1

gskril:
Also fixed a few broken code examples and inaccurate visuals (e.g. the primary name of 0x225...c3B5 is not luc.cash)


lucemans (approved):
LGTM

PR #363, opened 2024-11-27
Fix DAO stewards' avatars

gskril:
- Updated katherine's primary ENS name
- Switch to fetching avatars from wagmi instead of enstate (was returning 404s for many of them like [slobo.eth](https://enstate.rs/i/slobo.eth))
- Added `<Avatar>` component to handle fallbacks and reduce repeat code
- Fixed rule in prettier so my IDE would stop yelling at me

### DIFF
--- a/app/.prettierrc
+++ b/app/.prettierrc
@@ -1,5 +1,6 @@
 {
     "tabWidth": 4,
     "useTabs": false,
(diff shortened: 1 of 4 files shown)


Mask27aa (commented):
👍👍

PR #364, opened 2024-11-28
Added web3j code samples

NickSneo:
This PR adds code samples for Web3j library - https://github.com/hyperledger-web3j/web3j

### DIFF
--- a/docs/resolution/names.mdx
+++ b/docs/resolution/names.mdx
@@ -71,6 +71,11 @@ from namehash import namehash
 node = namehash('name.eth')
 ```
 
+```java {{ title: 'Web3j (Java)', variant: 'web3j', link: 'https://docs.web3j.io/4.11.0/advanced/ethereum_name_service/' }}
+byte[] nameHash = NameHash.nameHashAsBytes('luc.eth');
+String nameHashString = Numeric.toHexString(nameHash)
+```
+
 ```rust {{ title: "namehash-rust", link: "https://github.com/InstateDev/namehash-rust" }}
 fn main() {
   let node = &namehash("name.eth");
@@ -177,6 +182,12 @@ const labelhash = keccak256(toUtf8Bytes(normalizedLabel))
 string constant label = "label";
 bytes32 constant labelhash = keccak256(bytes(label));
 ```
+
+```java {{ title: 'Web3j (Java)', variant: 'web3j', link: 'https://docs.web3j.io/4.11.0/advanced/ethereum_name_service/' }}
+byte[] labelHash = NameHash.nameHashAsBytes('label');
+String labelHashString = Numeric.toHexString(labelHash)
+```
+
 </CodeGroup>
 
 ## DNS Encoding {{ title: "DNS Encoding", id: "dns" }}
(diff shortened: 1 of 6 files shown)


docs/resolution/names.mdx
```diff
@@ -177,6 +182,12 @@ const labelhash = keccak256(toUtf8Bytes(normalizedLabel))
 string constant label = "label";
 bytes32 constant labelhash = keccak256(bytes(label));
 ```
+
+```java {{ title: 'Web3j (Java)', variant: 'web3j', link: 'https://docs.web3j.io/4.11.0/advanced/ethereum_name_service/' }}
+byte[] labelHash = NameHash.nameHashAsBytes('label');
```
gskril, line 187: Labelhash is the keccak256 hash of the label (e.g. "nick" in the name "nick.eth"), which is different than applying `NameHash.nameHashAsBytes()` on the label.

Web3j likely doesn't have a helper function specific to labelhash, but it may have a keccak256 function which can be used in this example.

gskril (changes requested):
thanks for contributing and sorry for the review delay! just have one comment about labelhash

docs/resolution/names.mdx
NickSneo: So `NameHash.nameHash` Function can be used for both full Name like `nick.eth` and just label like 'nick`
It will give the correct result, thats why there is no ultitly function just for labels as user can use current `NameHash.nameHash` fucntion on label to get the labelHash.

This fuctions returns (Hash.sha3(result)) which is keccak256
Hope it makes sense, let me know if there is still any confusion

gskril: Hmmm I'm running a test script locally and `NameHash.nameHashAsBytes('label')` does not return the expected label hash. The expected value for `label` would be `0x1b036544434cea9770a413fd03e0fb240e1ccbd10a452f7dba85c8eca9ca3eda`

I don't know Java well, but futzing around it looks like the code example will have to be a bit lower level like this:

```
public static String labelHash(String name) {
      // Get the label of the name (the part before the earliest '.')
      String[] parts = name.split("\\.");
      String label = parts[0];

      // Keccak256 hash the label and convert to hex string
      byte[] labelHash = Hash.sha3(label.getBytes(StandardCharsets.UTF_8));
      String labelHashString = Numeric.toHexString(labelHash);
      return labelHashString;
    }
```

docs/resolution/names.mdx
```diff
@@ -71,6 +71,11 @@ from namehash import namehash
 node = namehash('name.eth')
 ```
 
+```java {{ title: 'Web3j (Java)', variant: 'web3j', link: 'https://docs.web3j.io/4.11.0/advanced/ethereum_name_service/' }}
+byte[] nameHash = NameHash.nameHashAsBytes('luc.eth');
+String nameHashString = Numeric.toHexString(nameHash)
```
gskril, line 76: Looks like web3j also exposes [`NameHash.nameHash`](https://github.com/hyperledger-web3j/web3j/blob/main/core/src/main/java/org/web3j/ens/NameHash.java#L37-L40) which returns a hex string directly, so no need to first get the bytes and convert it yourself

gskril (changes requested):
Sorry I missed this the first time — would be great if you can include the relevant imports for each example so they're truly copy/paste snippets!

docs/resolution/names.mdx
NickSneo: @gskril Thanks for pointing out this. I was able to reproduce it, and created a PR for labelHash fucntion - https://github.com/hyperledger-web3j/web3j/pull/2140

We will soon release new Web3j with this fix. Thanks

gskril:
The docs have been refactored quite a bit since this PR was opened, so I'm closing it for simplicity. It looks like the fix to the upstream library was merged, so I'd be happy to review a new PR if you get around to it!

PR #367, opened 2024-12-03
fix: min_commitment_age language

nftchance:
Spent several minutes chasing my tail until just going to look onchain and read the registrar logic. The minimum commitment age is enforced as a functional requirement to `commit`. Currently the docs mark it as "recommended" however that is not the case.


gskril (approved):
Lgtm, thanks for the clarification!

PR #371, opened 2024-12-10
Add local PDF of DAO investment policy

estmcmxci:
- updated protocol.ts to introduce new item to the directory 'Investment Policy Proposal'
- introduced 'ips.mdx' to include a local instance of the Investment Policy Proposal
- removed insecure, third-party links and linked back to a local instance of the IPS instead


gskril:
Sorry this was ignored. I agree we shouldn't rely on Google Drive links in the long term. I decided to host the PDF in this repo instead of copy the contents into a new doc for posterity. Thanks for contributing!

PR #379, opened 2024-12-18
Add [EP 5.28] to Governance Docs

5ajaki:
@gskril - this will still need the magical metadata.
We also need to determine how we're going to denote proposals that pass from those that fail.  Currently it's not represented in the governance docs at all.


gskril (approved):
Thanks for adding this! And noted — adding proposal status is on my list.

PR #381, opened 2024-12-25
Fix typos

rex4539:


gskril (approved):
Lgtm, thanks! [ensdomains/ensips](https://github.com/ensdomains/ensips) will soon become the source of truth for all ENSIP pages, so feel free to make the relevant changes there as well.

PR #404, opened 2025-03-10
feat: add ensip-20 to docs

pikonha:
Adds the ENSIP-20 Wildcard Writing to the docs UI.

Disclaimer: the mermaid sequence diagrams don't seem to be correctly handled by the UI, an raw diagram was added instead which decreases the readability of the flow.


Arachnid (approved):

gskril:
Thanks! I'm currently working on refactoring the docs to pull from the ensips repo directly which should properly handle the mermaid diagram, but this works great for now.

PR #421, opened 2025-04-22
Create 6.7.mdx

Arachnid:


gskril (approved):
The table is overflowing on some screens but that's a platform issue so I'll fix it in another PR. Otherwise looks good!

PR #448, opened 2025-07-17
Add EP 6.15

gskril:
This is an automated pull request to add a new DAO proposal to the ENS docs.

PR #449, opened 2025-07-17
Update EP 6.15

gskril:
This is an automated pull request to add a new DAO proposal to the ENS docs.


gskril:
This was posted a few minutes after the initial proposal (#448), so I'm keeping the same number even though it doesn't match Snapshot.

PR #454, opened 2025-08-07
Add Plausible

gskril:
Adds [Plausible](https://plausible.io/) for analytics. Puts their script and `/event` endpoint behind a proxy via Cloudflare Pages Functions [so that Adblockers don't block tracking](https://plausible.io/docs/proxy/introduction).

Also adds a "Did you find this page useful?" section to the bottom of every page, of which the responses get tracked as custom events in Plausible.

### DIFF
--- a/functions/api/blah/event.ts
+++ b/functions/api/blah/event.ts
@@ -1,0 +1,7 @@
+// Plausible Proxy
+// https://plausible.io/docs/proxy/guides/cloudflare
+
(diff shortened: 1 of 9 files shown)


schmidsi (approved):

functions/api/blah/event.ts
schmidsi, line 1: Why "blah"?

PR #542, opened 2026-04-20
Add ENS MCP by Namespace

thecaphimself:
Our ENS MCP is already live in production with 2 products - Web3Identity and ENScribe, and is steadily gaining traction (14 GitHub stars and growing). Given that it’s actively used and will be continuously maintained by the Namespace team, it would be great to include it in the official ENS documentation as a supported integration path.


thecaphimself:
Alternatively, for clarity purposes, instead of writing "ENS MCP" as it's written now, we can write "Namespace ENS MCP".

PR #543, opened 2026-04-27
V2 docs WIP

wentelteefje:


src/pages/contracts/ensv2/migration.mdx
```diff
@@ -0,0 +1,169 @@
...
+Locked names are those with the `CANNOT_UNWRAP` [fuse](/wrapper/fuses) burned in the ENSv1 [Name Wrapper](/wrapper/overview). These names have irrevocable restrictions that must be preserved in ENSv2.
+
+**How it works:**
+
+1. The name owner transfers their wrapped ERC1155 token to the `LockedMigrationController`
+2. The controller validates:
+   - The name is actually locked (`CANNOT_UNWRAP` is burned)
+   - The name data matches the expected node and owner
+3. The controller converts ENSv1 fuses to ENSv2 roles (see [Fuse-to-Role Conversion](#fuse-to-role-conversion))
+4. If the name has subnames, a `WrapperRegistry` is created for the child registry
```
hiddentao: Wrapper registry is always created, not just if it has subnames - actually we can't know this on-chain anyway

src/pages/contracts/ensv2/migration.mdx
```diff
@@ -0,0 +1,169 @@
...
+5. The name is registered as `RESERVED` in the ENSv2 registry with the converted role bitmap
```
hiddentao: The name is registered (RESERVED → REGISTERED) in the ENSv2 registry against its pre-existing reserved slot, using the converted role bitmap and inheriting the reserved expiry. This requires the slot to have been moved to RESERVED beforehand by pre-migration.

src/pages/contracts/ensv2/migration.mdx
```diff
@@ -0,0 +1,169 @@
...
+### Unlocked and Unwrapped Names (UnlockedMigrationController)
+
+The `UnlockedMigrationController` handles two types of names:
+
+**Unlocked wrapped names** (NameWrapper ERC1155 tokens without `CANNOT_UNWRAP`):
+
+1. The owner transfers their wrapped ERC1155 token to the controller
+2. The controller validates the name is NOT locked
+3. The controller clears the ENSv1 resolver
+4. The name is registered as `RESERVED` in ENSv2
```
hiddentao: No, reservations take place during the pre-migration phase.

src/pages/contracts/ensv2/migration.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+**Unwrapped names** (BaseRegistrar ERC721 tokens):
+
+1. The owner transfers their ERC721 token to the controller via `safeTransferFrom`
+2. The controller's `onERC721Received` handler processes the migration
+3. The name is registered as `RESERVED` in ENSv2
```
hiddentao: Again, same issue

src/pages/contracts/ensv2/migration.mdx
```diff
@@ -0,0 +1,169 @@
...
+### Locked Names (LockedMigrationController)
+
+Locked names are those with the `CANNOT_UNWRAP` [fuse](/wrapper/fuses) burned in the ENSv1 [Name Wrapper](/wrapper/overview). These names have irrevocable restrictions that must be preserved in ENSv2.
+
+**How it works:**
+
+1. The name owner transfers their wrapped ERC1155 token to the `LockedMigrationController`
+2. The controller validates:
+   - The name is actually locked (`CANNOT_UNWRAP` is burned)
+   - The name data matches the expected node and owner
```
hiddentao: Bit misleading. Premigration sets the expiry and migration supplies the new owner and subregisty address. Note that new owner does not need to equal owner in v1

src/pages/contracts/ensv2/migration.mdx
```diff
@@ -0,0 +1,169 @@
...
+    end
+
+    locked -->|"ERC1155 transfer"| lmc
+    unlocked -->|"ERC1155 transfer"| umc
+    unwrapped -->|"ERC721 transfer"| umc
+    lmc -->|"register as RESERVED"| registry
+    umc -->|"register as RESERVED"| registry
+```
+
+All migration paths result in the name being registered as `RESERVED` in the ENSv2 [Permissioned Registry](/contracts/ensv2/permissioned-registry), preserving the existing expiry. The name is then promoted to `REGISTERED` when the owner claims it.
```
hiddentao: Pre-migration reserves names. Migrations registers them, i.e. promoted the pre-migrated status to registered.

In the reserved phase names have an expiry matching what's in v1 (+ 90 days to cover grace period) and resolve using the v1 resolver. But they have no owner set.

Migrations sets the owner.

src/pages/contracts/ensv2/migration.mdx
```diff
@@ -0,0 +1,169 @@
...
+| ENSv1 Fuse              | ENSv2 Equivalent                              | Notes                                    |
+| ----------------------- | --------------------------------------------- | ---------------------------------------- |
+| `PARENT_CANNOT_CONTROL` | Revoke parent's admin roles on the child name | Parent loses ability to modify the child |
+| `CAN_EXTEND_EXPIRY`     | Grant `ROLE_RENEW` on the child name          | Child owner can extend their own expiry  |
+
+The conversion is handled by `_subregistryRoleBitmapFromFuses()` and `_tokenRoleBitmapFromFuses()` in the migration contracts.
+
+## WrapperRegistry
+
+The **WrapperRegistry** acts as a bridge between ENSv1's [Name Wrapper](/wrapper/overview) and ENSv2's registry system. It wraps the Name Wrapper as an ENSv2-compatible registry, enabling a gradual migration where:
```
hiddentao: Bridge is misleading. Better to say it replicates the fuse-based access control in V2 using the V2 access control system.

src/pages/contracts/ensv2/migration.mdx
```diff
@@ -0,0 +1,169 @@
...
+| `CANNOT_SET_TTL`          | N/A                              | TTL is not stored in ENSv2 registries                              |
+
+### Parent-Controlled Fuses
+
+| ENSv1 Fuse              | ENSv2 Equivalent                              | Notes                                    |
+| ----------------------- | --------------------------------------------- | ---------------------------------------- |
+| `PARENT_CANNOT_CONTROL` | Revoke parent's admin roles on the child name | Parent loses ability to modify the child |
+| `CAN_EXTEND_EXPIRY`     | Grant `ROLE_RENEW` on the child name          | Child owner can extend their own expiry  |
+
+The conversion is handled by `_subregistryRoleBitmapFromFuses()` and `_tokenRoleBitmapFromFuses()` in the migration contracts.
```
hiddentao: _subregistryRoleBitmapFromFuses unconditionally adds ROLE_RENEW | ROLE_RENEW_ADMIN for the subregistry independent of any fuse. 

Also, admin counterparts are added by the roleBitmap << 128 shift only when fuses are not frozen — both conversion functions do this

hiddentao (changes requested):
Migration docs need a lot of work

src/pages/contracts/ensv2/migration.mdx
wentelteefje: Addressed in new migration page (WIP).

src/pages/contracts/ensv2/migration.mdx
wentelteefje: Addressed in new migration page (WIP).

src/pages/contracts/ensv2/migration.mdx
wentelteefje: Addressed in new migration page (WIP).

src/pages/contracts/ensv2/migration.mdx
wentelteefje: Addressed in new migration page (WIP).

src/pages/contracts/ensv2/migration.mdx
wentelteefje: Addressed in new migration page (WIP).

src/pages/contracts/ensv2/migration.mdx
wentelteefje: Addressed in new migration page (WIP).

src/pages/contracts/ensv2/overview.mdx
```diff
@@ -42,161 +44,69 @@ The information on this page is a work-in-progress. Expect updates as the design
...
+| [Mutable Token IDs](/contracts/ensv2/mutable-token-ids)             | Dynamic token IDs that protect against permission leaks and griefing         |
+| [Hidden Contract Accounts](/contracts/ensv2/hca)                    | Smart-account attribution for HCA proxy wallets                              |
+| [Verifiable Factory](/contracts/ensv2/verifiable-factory)           | Deterministic CREATE2 proxy deployment with on-chain verification            |
+
+## Contracts
+
+| Page                                                            | Description                                                        |
+| --------------------------------------------------------------- | ------------------------------------------------------------------ |
+| [Permissioned Registry](/contracts/ensv2/permissioned-registry) | Tokenized registry managing name ownership, state, and permissions |
+| [Permissioned Resolver](/contracts/ensv2/permissioned-resolver) | Per-name resolver with per-record roles and record aliasing        |
```
makoto: `Per-name resolver` is incorrect. Please refer to https://github.com/ensdomains/contracts-v2/blob/b682ca4c0d47590b6ecce435399d89c4498d5d8f/docs/indexing-ensv2-events.md?plain=1#L23

src/pages/contracts/ensv2/overview.mdx
```diff
@@ -42,161 +44,69 @@ The information on this page is a work-in-progress. Expect updates as the design
...
-| [Registries](/contracts/ensv2/registries) | Registries govern ownership of names                    |
-| [Resolvers](/contracts/ensv2/resolvers)   | Resolvers control how a name is converted to a resource |
-
-## Upgrade timeline
-
-The team is targeting **Q4 2024** for mainnet deployment, subject to audit and
-community review. A detailed upgrade guide will be published here soon. */}
+## Registry Architecture
+
+ENSv2 replaces ENSv1's single flat registry with a **hierarchical model** where each name can have its own registry for managing subnames, and its own resolver for storing records. Names don't exist as single on-chain objects. Instead, a full name like `sub.alice.eth` is a chain of entries across registries linked by subregistry pointers.
```
makoto: `its own resolver for storing records.` is incorrect as stated elsewhere

src/pages/contracts/ensv2/dns-resolvers.mdx
```diff
@@ -0,0 +1,92 @@
...
+
+<FrenCallout fren="lili" variant="tip">
+The contracts and interfaces described here are **not yet final** and may change prior to mainnet deployment.
+</FrenCallout>
+
+## How It Works
+
+When the Universal Resolver encounters a DNS name (e.g., `example.com`), it finds the **DNSTLDResolver** set as the resolver for that TLD on the root registry. The DNSTLDResolver then follows a multi-step resolution strategy:
+
+1. **Check ENSv1**: look for an existing resolver in the ENSv1 registry. If one exists (and it's not the old v1 DNS resolver), delegate to it directly. This preserves backward compatibility.
```
makoto: You may want to mention that onchain dnssec resolution remains in v1 at the point of the v2 launch (and explain the difference between the onchain and offchain dnssec version)

src/pages/contracts/ensv2/eth-registrar.mdx
```diff
@@ -0,0 +1,487 @@
...
+
+<FrenCallout fren="lili" variant="tip">
+The contracts and interfaces described here are **not yet final** and may change prior to mainnet deployment.
+</FrenCallout>
+
+## What Changed from ENSv1
+
+| Feature              | ENSv1                               | ENSv2                                                                                 |
+| -------------------- | ----------------------------------- | ------------------------------------------------------------------------------------- |
+| Grace period         | 90-day grace period after expiry    | **No grace period**, names expire immediately                                        |
```
makoto: grace period came back (28 days)

src/pages/contracts/ensv2/overview.mdx
gskril: I think this was written for Namechain. Should be removed/revised

gskril: I think this was changed after removing Namechain for gas savings - each user/account gets its own resolver now, not every name

src/pages/ensv2/overview.mdx
```diff
@@ -42,161 +44,69 @@ The information on this page is a work-in-progress. Expect updates as the design
...
+| --------------------------------------------------------------- | ------------------------------------------------------------------ |
+| [Permissioned Registry](/contracts/ensv2/permissioned-registry) | Tokenized registry managing name ownership, state, and permissions |
+| [Permissioned Resolver](/contracts/ensv2/permissioned-resolver) | Per-name resolver with per-record roles and record aliasing        |
+| [ETH Registrar](/contracts/ensv2/eth-registrar)                 | Commit-reveal registration and renewal for .eth names              |
+| [Universal Resolver V2](/contracts/ensv2/universal-resolver-v2) | Single entry point for name resolution across the hierarchy        |
+| [Registry Metadata](/contracts/ensv2/registry-metadata)         | ERC-1155 metadata providers for registries                         |
+| [DNS Name Resolution](/contracts/ensv2/dns-resolvers)           | Resolving DNS domain names through the ENS protocol                |
+| [Reverse Resolution](/contracts/ensv2/reverse-resolution)       | Primary names and multi-chain reverse resolution                   |
+
+## Deployments (Sepolia)
```
gskril, line 113: Not necessary to move this, but FYI the [deployments page](https://a7d09a51.docs-bao.pages.dev/learn/deployments/) is auto generated from the ensdomains/ens-contracts repo which might be useful so we don't have to maintain multiple lists

src/components/ensv2/IdExplorer.tsx
gskril, line 1: The sliders don't seem to change any of the values for me 

Image

src/pages/ensv2/indexing.mdx
```diff
@@ -0,0 +1,500 @@
...
+```
+
+Each registry is a [Permissioned Registry](/contracts/ensv2/permissioned-registry) (or UserRegistry for subnames), implementing `IRegistry`, `IStandardRegistry`, `IPermissionedRegistry`, and `IEnhancedAccessControl`. Tokens are [ERC1155Singleton](/contracts/ensv2/erc1155-singleton) (one token per name).
+
+Resolvers are separate contracts ([Permissioned Resolver](/contracts/ensv2/permissioned-resolver)) deployed per-name as UUPS proxies. Multiple names can share the same resolver via [aliases](/contracts/ensv2/permissioned-resolver#aliasing) (`setAlias`), which allow one name to reuse another's resolver records by rewriting the name suffix during resolution (e.g., `sub.alias.eth` → `sub.test.eth`).
+
+## Registry Events
+
+These events are emitted by any registry contract (PermissionedRegistry / UserRegistry).
+
```
makoto, line 29: Please add TODO section stating that `RegistryCreated` is due to be included https://github.com/ensdomains/contracts-v2/pull/258

src/pages/contracts/ensv2/mutable-token-ids.mdx
```diff
@@ -0,0 +1,187 @@
...
+| First registration             |      -       |       -        |       New token minted       | Both counters start at 0 (storage default)                  |
+| Role grant/revoke              |      -       |       +1       |   Yes (regeneration)         | Invalidates marketplace approvals                           |
+| Transfer                       |      -       |       -        |              No              | Owner changes, same token ID                                |
+| Re-registration (after expiry) |      +1      |       +1       | Yes (old burned, new minted) | Old token burned, all prior role grants invalidated         |
+| Unregistration (registered)    |      +1      |       +1       |         Yes (burned)         | Token destroyed, versions incremented for next registration |
+| Unregistration (reserved)      |      -       |       -        |             No               | No token exists to burn; counters unchanged                 |
+| Renewal                        |      -       |       -        |              No              | Only expiry changes                                         |
+
+## anyId Polymorphism
```
gskril:
```suggestion
## `anyId` Polymorphism
```

src/pages/contracts/ensv2/mutable-token-ids.mdx
```diff
@@ -0,0 +1,187 @@
...
+    style BUILD_TK fill:#F5F5F5,stroke:#0080BC,color:#011A25
+    style BUILD_RS fill:#F5F5F5,stroke:#0080BC,color:#011A25
+    style USE_TK fill:#0080BC,stroke:#011A25,color:#fff
+    style USE_RS fill:#0080BC,stroke:#011A25,color:#fff
+```
+
+Functions that accept `anyId` include: `setSubregistry()`, `setResolver()`, `renew()`, `unregister()`, `getExpiry()`, `getStatus()`, `getState()`, `getTokenId()`, `getResource()`, and all EAC role functions.
+
+:::note
+anyId polymorphism is specific to the [Permissioned Registry](/contracts/ensv2/permissioned-registry). The [Permissioned Resolver](/contracts/ensv2/permissioned-resolver) uses a [different resource scheme](/contracts/ensv2/enhanced-access-control#resources) based on `keccak256(node, part)` and does not use anyId.
```
gskril:
```suggestion
`anyId` polymorphism is specific to the [Permissioned Registry](/contracts/ensv2/permissioned-registry). The [Permissioned Resolver](/contracts/ensv2/permissioned-resolver) uses a [different resource scheme](/contracts/ensv2/enhanced-access-control#resources) based on `keccak256(node, part)` and does not use anyId.
```

src/pages/contracts/ensv2/mutable-token-ids.mdx
```diff
@@ -0,0 +1,187 @@
...
+:::
+
+## Implications for Developers
+
+### Don't Cache Token IDs
+
+Token IDs are not stable identifiers. If you need to reference a name, store the **labelhash** instead and use `getTokenId(anyId)` to get the current token ID when needed.
+
+### Use anyId Where Possible
```
gskril:
```suggestion
### Use `anyId` Where Possible
```

Maybe its just me, but "anyId" looks weird as normal text so i feel like should always be formatted as code

src/pages/ensv2/verifiable-factory.mdx
gskril, line 1: Resolvers were changed to be per-account vs per-name when we moved back to L1, which would change a decent amount of content on this doc

Also maybe worth calling out that custom resolvers are still allowed at any level (per-acocunt, per-name, whatever) and the factory is just our default implementation to replace the Public Resolver

src/pages/contracts/ensv2/eth-registrar.mdx
```diff
@@ -0,0 +1,487 @@
...
+    label string,        // The label to renew
+    duration uint64,     // Duration to extend by (in seconds)
+    paymentToken IERC20, // ERC20 token to pay with (must be approved)
+    referrer bytes32     // Referral identifier
+)
+```
+
+## BatchRegistrar
+
+The **BatchRegistrar** is a companion contract used during migration. It allows the deployer to batch-register or batch-renew names in a single transaction:
```
gskril:
```suggestion
The **BatchRegistrar** is a companion contract used during migration from ENSv1. It allows the deployer to batch-register or batch-renew names in a single transaction:
```

src/pages/contracts/ensv2/universal-resolver-v2.mdx
```diff
@@ -0,0 +1,237 @@
...
+</FrenCallout>
+
+## Architecture
+
+`UniversalResolverV2` extends `AbstractUniversalResolver`, a shared base contract that implements all resolution logic: forward resolution, reverse resolution, [CCIP-Read](/resolvers/ccip-read) gateway batching, and callback chaining. The only function the base leaves abstract is `findResolver()`, which each version overrides with its own registry traversal strategy:
+
+- **URv1** walks a flat ENS registry
+- **URv2** walks the hierarchical registry tree via `getSubregistry()` at each level
+
+This means `resolve()`, `reverse()`, and all CCIP-Read infrastructure work identically across both versions. The `LibRegistry` library that powers the hierarchy traversal is also used directly by other contracts including `DNSAliasResolver`, `DNSTLDResolver`, and `ENSV2Resolver`.
```
gskril: Raffy explained this nicely on a call once that stuck with me and might be worth including in some form. In v1, resolution was bottom-up whereas v2 is top-down.

Used to:
1. Look for resolver of sub.sub.name.eth
2. No resolver is set, so try sub.name.eth
3. No resolver is set, so try name.eth

Now its:
1. Get resolver of eth
2. Get resolver of name.eth
3. No resolver for sub.name.eth, use the one from step 2

This is what allows names to stop resolving after expiration and some other similar fine grained behaviors.

src/pages/ensv2/universal-resolver-v2.mdx
```diff
@@ -0,0 +1,237 @@
...
+2. If non-zero, query the registry for the name's token ownership
+
+The bidirectional check ensures that:
+- The registry is actually reachable from the root (downward walk)
+- The registry considers itself to be at that position in the hierarchy (upward walk via `getParent()`)
+- No aliasing or stale references are being exploited
+
+## Reference
+
+### View Functions
```
gskril, line 276: Missing `requireResolver()` which i understand to be the safe version of `findResolver()` and what's actually used internally

src/pages/contracts/ensv2/reverse-resolution.mdx
```diff
@@ -0,0 +1,118 @@
+---
+description: Primary names and reverse resolution in ENSv2, including multi-chain support and signature-based claims
+---
+
+import { FrenCallout } from '../../../components/ensv2/FrenCallout'
+
+# Reverse Resolution
+
+Reverse resolution maps an address back to an ENS name (the "primary name"). In ENSv1, this was handled by a single L1 Reverse Registrar. ENSv2 redesigns reverse resolution with native multi-chain support, signature-based claims, and a standalone architecture detached from the registry hierarchy.
```
gskril: v1 has multi-chain reverse registrars and signature-based claims https://docs.ens.domains/web/reverse#l2-primary-names

gskril (commented):
Looking great! Left some comments/suggestions

src/components/ensv2/IdExplorer.tsx
wentelteefje: They even increment in your animation :D 

- `tokenVersionId` increments the lower 32 bits of Token ID.
- `eacVersionId` increments the lower 32 bits of resource. 

The canonicalId always zeroes out the lower 32 bits, so nothing changes there and the labelhash is just a constant 256bit hash. But it's good feedback that you didn't even realize them changing. Maybe I should add some visual effect to the values when the slider is used.

gskril: oh whoops, ya i was looking at the left side 😅

seems maybe weird to display the same value 4 times when only the right side changes, but not a big deal for now

src/pages/ensv2/migration.mdx
```diff
@@ -0,0 +1,288 @@
...
+Unlocked 2LD names (NameWrapper ERC-1155 tokens without `CANNOT_UNWRAP`) are also migrated via the `UnlockedMigrationController`.
+
+1. Transfer the ERC-1155 token via `safeTransferFrom` (or batch via `safeBatchTransferFrom`) with an encoded data payload
+2. The controller verifies the name is NOT locked (reverts `NameIsLocked` otherwise)
+3. The token is unwrapped to the Graveyard and the v1 resolver is cleared
+4. The name is promoted from RESERVED to REGISTERED with the provided parameters
+5. Token roles are the same as `ETHRegistrar.register()`
+
+:::note
+Unlocked 3LD+ names cannot be migrated through this path. They must be registered directly in v2 once their parent has a v2 registry.
```
makoto, line 142: @hiddentao can you check if this is true?

src/pages/contracts/ensv2/overview.mdx
makoto: is this true? what part was extended to call it `native l2 support`? I am wondering this is left over of the time when v2 was going to be deployed to own l2 (aka namechain)

src/pages/contracts/ensv2/migration.mdx
```diff
@@ -0,0 +1,288 @@
...
+
+Names that *could* migrate but don't will eventually expire. Their v1 tokens expire, leaving the v1 registry state frozen (resolver, owner, fuses all unchanged). The v2 RESERVED slot also expires. Once both have expired, the name becomes available for fresh registration in v2 via the `ETHRegistrar`.
+
+## Scenarios
+
+### 100 days remaining
+
+*A name with 100 days left in v1 is premigrated.*
+
+The RESERVED slot has 162 days left (100 + 62 day bonus). ENSv1 expires after 100 days and becomes available after 190 days (100 + 90 day grace). ENSv2 expires after 162 days and becomes available after 190 days (162 + 28 day grace). Both systems make the name available at the same time.
```
hiddentao: Note that names do not become availalbe in v1 again once v2 is live and user migration is possible because we will disable v1 registrations and renewals entirely. In v1 names will simply expire and no longer resolve because we'll send them to the Graveyard. 

_Note: this comment applies to all other instances where v1 availability is mentioned_.

src/pages/contracts/ensv2/migration.mdx
```diff
@@ -0,0 +1,288 @@
...
+
+:::note
+Migration is **not required** immediately. Unmigrated names continue to resolve through v1 via `ENSV1Resolver`. However, migrating gives you access to v2 features like [per-record permissions](/contracts/ensv2/permissioned-resolver#eac-integration), [aliasing](/contracts/ensv2/permissioned-resolver#aliasing), and the new resolver.
+:::
+
+### Subname Owners
+
+If you own a locked subname (3LD+), your parent must migrate first. Once the parent has migrated, transfer your token to the parent's `WrapperRegistry`.
+
+Unlocked 3LD+ subnames cannot be migrated through the migration controllers. They must be registered directly in v2 once their parent has a v2 registry.
```
hiddentao: Not quite true. Emancipated (i.e. `PARENT_CANNOT_CONTROL` fuse is set but `CANNOT_UNWRAP` fuse is NOT set) names can be migrated through the migration controllers - they simply get unwrapped and registered in the parent's `WrapperRegistry`.

hiddentao (changes requested):

src/pages/ensv2/permissioned-registry.mdx
```diff
@@ -0,0 +1,562 @@
...
+stateDiagram-v2
+    [*] --> AVAILABLE
+    AVAILABLE --> REGISTERED: register()
+    AVAILABLE --> RESERVED: register() with owner=0
+    RESERVED --> REGISTERED: register()
+    REGISTERED --> AVAILABLE: unregister() or expiry
+    RESERVED --> AVAILABLE: unregister() or expiry
+```
+
+- `AVAILABLE`: never registered or expired. Open for registration.
```
makoto, line 61: is it true? Once expired, what stage is it going to be called?

src/pages/contracts/ensv2/permissioned-resolver.mdx
```diff
@@ -0,0 +1,738 @@
...
+
+import { Card } from '../../../components/ui/Card'
+import { FrenCallout } from '../../../components/ensv2/FrenCallout'
+import { ResourceCalculator } from '../../../components/ensv2/ResourceCalculator'
+import { RoleBitmapComposer } from '../../../components/ensv2/RoleBitmapComposer'
+import { ContractReference } from '../../../components/ensv2/ContractReference'
+
+# Permissioned Resolver
+
+In ENSv1, most names shared a single [Public Resolver](/resolvers/public) contract. In ENSv2, **each name gets its own resolver instance**, deployed as a UUPS-upgradeable proxy. This gives name owners fine-grained control over who can set which records, and enables new features like record aliasing.
```
makoto: each user, not each name.

src/pages/ensv2/verifiable-factory.mdx
```diff
@@ -0,0 +1,132 @@
+---
+description: Verifiable Factory
+---
+
+import { FrenCallout } from '../../../components/ensv2/FrenCallout'
+
+# Verifiable Factory
+
+ENSv2 deploys per-name resolvers and per-name subname registries through a single shared factory. Each instance is a UUPS proxy with a deterministic CREATE2 address and an on-chain proof of provenance. This page covers the factory itself and how the protocol uses it.
+
```
makoto, line 12: Verifiable Factory is in the separate repo. Should have a link to https://github.com/ensdomains/verifiable-factory

src/pages/ensv2/hca.mdx
```diff
@@ -0,0 +1,85 @@
+---
+description: Hidden Contract Accounts
+---
+
+import { FrenCallout } from '../../../components/ensv2/FrenCallout'
+
+# Hidden Contract Accounts
+
+Hidden Contract Accounts (HCAs) let smart-account proxies act on behalf of their owners while still being attributed to the owner. ENSv2 wires HCA-aware sender resolution into every layer of the stack. When a user controls a name through an HCA proxy, the protocol records the owner as the actor for permission checks, ownership reads, and event indexing.
+
```
makoto: Even though v2 doesn't have hard dependency to HCA, it should be worth mentioned that it will be used in combination with https://github.com/rhinestone-external/ens-modules and have some code example

makoto:
Hi, I already made comments to individual lines. These are more generic comments to the part it wasn't changed on the PR

- `No Grace period` mentioned [here](https://github.com/ensdomains/docs/pull/543/changes#diff-ad3359fe48e2f813a3c72673bc1a935b457fb718a700d41cfab4dd471777a885R29 ) is no longer correct.
- `every name gets its own resolver,` mentioned [here](https://github.com/ensdomains/docs/pull/543/changes#diff-ad3359fe48e2f813a3c72673bc1a935b457fb718a700d41cfab4dd471777a885R31) is incorrect. While each name has its own registry, name is per user. Please refer to [here](https://github.com/ensdomains/contracts-v2/blob/b682ca4c0d47590b6ecce435399d89c4498d5d8f/docs/indexing-ensv2-events.md?plain=1#L23) for more detail
- `all existing ENSv1 resolvers will continue to work seamlessly in ENSv2.` is technically incorrect. The migration contract takes a resolver address to set the resolver in v2 so it's up to the client app. For ENS app, new v2 permissioned resolver will be created and data is migrated to the new resolver at the time of the migration, but please confirm with the frontend team about the exact behaviour.
- `hca.mdx` is a bit abstract. There should be a code example section (when hca component is ready). it's also worth noting who is in responsible for creating HCA at what timing as ENS v2 protocol itself doesn't have the HCA creation functionality but it expects to use a single hca factory address across all the HCA wired contracts.

src/pages/contracts/ensv2/overview.mdx
wentelteefje: Fixed.

src/pages/contracts/ensv2/overview.mdx
wentelteefje: Fixed.

src/pages/contracts/ensv2/eth-registrar.mdx
wentelteefje: Fixed.

src/pages/contracts/ensv2/permissioned-resolver.mdx
wentelteefje: Fixed.

src/pages/contracts/ensv2/migration.mdx
wentelteefje: Fixed.

src/pages/contracts/ensv2/migration.mdx
wentelteefje: Fixed.

src/pages/contracts/ensv2/overview.mdx
wentelteefje: Removed.

wentelteefje: Fixed.

wentelteefje: Yes, left-over that I didn't remove yet. Fixed.

src/pages/contracts/ensv2/dns-resolvers.mdx
wentelteefje: Fixed.

src/pages/ensv2/indexing.mdx
wentelteefje: Done.

src/pages/ensv2/universal-resolver-v2.mdx
wentelteefje: Added.

src/pages/ensv2/permissioned-registry.mdx
wentelteefje: An expired name will be in the `AVAILABLE` state. See https://github.com/ensdomains/contracts-v2/blob/main/contracts/src/registry/PermissionedRegistry.sol#L570-L572

src/pages/contracts/ensv2/eth-registrar.mdx
wentelteefje: Fixed

src/pages/contracts/ensv2/mutable-token-ids.mdx
wentelteefje: Fixed

src/pages/contracts/ensv2/mutable-token-ids.mdx
wentelteefje: Fixed

src/pages/ensv2/verifiable-factory.mdx
wentelteefje: Thanks for pointing that out. Added.

src/pages/contracts/ensv2/reverse-resolution.mdx
wentelteefje: Whoops. fixed!

src/pages/ensv2/verifiable-factory.mdx
wentelteefje: Added

src/pages/contracts/ensv2/migration.mdx
```diff
@@ -0,0 +1,296 @@
...
+| **Locked** | NameWrapper ERC-1155 | Emancipated + `CANNOT_UNWRAP` (2LD and 3LD+) |
+| **Detached** | NameWrapper ERC-1155 | Emancipated, without `CANNOT_UNWRAP`, parent is Locked (3LD+ only) |
+
+Both Locked and Detached are subtypes of **Emancipated** (`PARENT_CANNOT_CONTROL` set). The distinction determines the migration path: Locked names preserve their fuse restrictions as v2 roles, while Detached names are unwrapped and registered with the standard role set.
+
+### Key Definitions
+
+- **RESERVED**: a v2 registry slot with an expiry and resolver but no owner. Created during premigration.
+- **REGISTERED**: a v2 registry slot with an owner. Created when a user migrates their v1 token.
+- **Graveyard**: the burn address for migrated v1 tokens. Also cleans up stale v1 registry entries.
```
hiddentao: Graveyard is not a burn address but a controller which contains functionality to burn expired v1 names such that they no longer resolve.

src/pages/contracts/ensv2/migration.mdx
```diff
@@ -0,0 +1,296 @@
...
+
+| ENSv1 Fuse | ENSv2 Role | Granted when fuse is... |
+|------------|-----------|------------------------|
+| `CANNOT_CREATE_SUBDOMAIN` | `ROLE_REGISTRAR` | Not set (not burned) |
+| `CANNOT_BURN_FUSES` | Admin counterparts of above | Not set (not frozen) |
+| (always) | `ROLE_RENEW` + `ROLE_RENEW_ADMIN` + `ROLE_UPGRADE` + `ROLE_UPGRADE_ADMIN` | Always granted |
+
+## Graveyard
+
+The `Graveyard` contract is the burn address for migrated v1 tokens. It holds migrated tokens until expiry. The tokens are technically transferable, but the Graveyard has no upgrade or transfer mechanism.
```
hiddentao: Graveyard is not a burn address but a controller which contains functionality to burn expired v1 names such that they no longer resolve.

src/pages/contracts/ensv2/migration.mdx
```diff
@@ -0,0 +1,296 @@
...
+
+- Resolution continues via `ENSV1Resolver` as long as the v2 reservation is active
+- Renewals work via `ETHRenewerV1`, keeping the v1 and v2 expiries in sync
+- v1 fuses, resolver, and ownership remain unchanged
+
+The only v2 features unavailable to unmigratable names are per-record permissions, aliasing, and the new resolver. Once the v2 reservation expires, the name becomes available for fresh registration in v2. The `Graveyard` can clear the expired v1 namespace.
+
+### What Happens to Unmigrated Names
+
+Names that *could* migrate but don't will eventually expire. Their v1 tokens expire, leaving the v1 registry state frozen (resolver, owner, fuses all unchanged). Since v1 registrations are disabled, names never become available in v1 again. The v2 RESERVED slot also expires. Once expired in v2, the name becomes available for fresh registration via the `ETHRegistrar`.
```
hiddentao: Actually our intention is to have the graveyarding daemon (Morticia) re-register/transfer these names to the graveyard so that they can be permanently burnt resolver-wise.

src/pages/contracts/ensv2/permissioned-registry.mdx
```diff
@@ -0,0 +1,738 @@
...
+You can only grant roles for which you hold the corresponding admin role. The set of roles granted at registration determines what the owner can delegate; see [Emancipation](#emancipation) for the default bitmap.
+</FrenCallout>
+
+### EAC Hook Overrides
+
+The Permissioned Registry overrides several [EAC callback hooks](/contracts/ensv2/enhanced-access-control#callback-hooks) to enforce registry-specific invariants:
+
+**Token regeneration on role changes**: when roles are granted or revoked via `grantRoles()` / `revokeRoles()`, the `_onRolesGranted` and `_onRolesRevoked` hooks trigger a token regeneration (burn + mint with a new token ID). This invalidates any pending ERC1155 transfer approvals tied to the old token ID, preventing an attacker from racing to transfer a token after their roles have been revoked.
+
+**Admin role restriction on names**: `_getSettableRoles` is overridden so that admin roles on individual names can only be assigned at registration time. After registration, only regular (non-admin) roles can be granted on a name. On `ROOT_RESOURCE`, admin roles work normally. This prevents a name owner from escalating their own permissions after registration.
```
hiddentao: Note that admin roles can still be revoked, including from oneself.

src/pages/contracts/ensv2/enhanced-access-control.mdx
```diff
@@ -0,0 +1,287 @@
+---
+description: Enhanced Access Control
+---
+
+import { FrenCallout } from '../../../components/ensv2/FrenCallout'
+import { ContractReference } from '../../../components/ensv2/ContractReference'
+
+# Enhanced Access Control
+
+Enhanced Access Control (EAC) is the permission system used throughout ENSv2. It controls who is allowed to do what, and on which names. EAC supports up to 2^256 independent resources (e.g., individual names), 64 roles per resource (32 regular + 32 admin), and up to 15 accounts per role.
```
hiddentao: `upto 15 holders of a role within the context of a single resource` to be more precise.

src/pages/contracts/ensv2/dns-resolvers.mdx
```diff
@@ -0,0 +1,97 @@
...
+
+Set the resolver to the DNSAliasResolver address and provide a rewrite rule in the context. Two modes are supported:
+
+**Suffix replacement**: rewrite the DNS suffix to an ENS suffix.
+
+```
+ENS1 dnsalias.ens.eth com base.eth
+```
+
+This causes `sub.example.com` to resolve as `sub.example.base.eth`. The resolver rewrites the suffix, then resolves the rewritten name through the ENSv2 registry hierarchy.
```
gskril: Where is `sub.example.com` introduced? I don't really understand this aliasing format so may be misunderstanding

wentelteefje: Thanks for catching that! Fixed.

src/pages/ensv2/indexing.mdx
```diff
@@ -0,0 +1,542 @@
...
+Children registered in `0xABC` appear under both parents. The token `wallet` in registry `0xABC` is simultaneously `wallet.sub1.sub2.parent.eth` and `wallet.linked.parent.eth` - they share the same tokenId.
+
+### Alias Resolution
+
+Aliases are a resolver-level concept, not a registry-level one:
+
+- `alias.eth` and `test.eth` may share the same resolver
+- The resolver stores an `alias.eth → test.eth` alias mapping
+- When resolving `sub.alias.eth`, the resolver rewrites it to `sub.test.eth` and returns those records
```
schmidsi, line 426: Is "rewrite" the correct word here? I understand "write" and "rewrite" as a state changing action, while this is just reading, no? So maybe "the resolver delegates it to ..."? or forwards

wentelteefje: It's a valid point, but I'm using the terminology introduced by the `DNSAliasResolver` [itself](https://github.com/ensdomains/contracts-v2/blob/main/contracts/src/dns/DNSAliasResolver.sol#L19-#L31). It's not state-changing in the sense that records get overwritten, but instead during alias resolution the resolution path is rewritten through "in-place" replacement (e.g. replace `alias.eth` by `test.eth`).

PR #546, opened 2026-05-06
Support ENSIP subdirectory files in docs generation

wentelteefje:
ENSIPs can now outsource supplementary content (e.g. tables) to markdown files in subdirectories (see https://github.com/ensdomains/ensips/pull/70)

Two changes:
- Filter GitHub API response to skip directory entries
- Inline subdirectory file content when an ENSIP links to it via a relative path like [text](./19/supported.md)


gskril (approved):
Lgtm

PR #547, opened 2026-05-08
Improve ENSIP generation reliability

wentelteefje:
- Use sidebar JSON (written last) as the completion marker instead of `1.mdx` (written first). If a previous run crashed midway, partial `.mdx` files won't prevent re-generation on the next build.
- Add `fetchWithRetry` wrapper for transient network failures (3 retries with progressive delay). 

TODO:
- [x] Add GITHUB_TOKEN for GitHub API call (api.github.com)

### DIFF
--- a/scripts/ensips.ts
+++ b/scripts/ensips.ts
@@ -27,8 +27,14 @@ export async function ensips() {
 
     console.log('Fetching ENSIPs')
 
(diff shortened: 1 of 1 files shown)


gskril (approved):
Lgtm

PR #557, opened 2026-06-19
Mark ethers.js v6.17.0 as ENSv2-compatible

schmidsi:


gskril:
Oops I already updated the page without seeing this PR https://github.com/ensdomains/docs/commit/e62cd85fb506cb95255d92c0d8b70338fc31580a

